PR #9, opened 2020-04-12
Add Travis integration

### LINKED ISSUE #5: Do not use constexpr for log_factorial
We cannot be sure how long the array should be anyway, so any `const`-ness would eventually lead to segfault.

### LINKED ISSUE #7: Set up CI
We have a Dockerfile and a functioning conan setup, all we need to do is run these somewhere.

lukipuki:
The usage of const is still a hack, we need a proper solution https://github.com/lukipuki/COVID-19-simulation/issues/5


miskosz (approved):
LGTM. Not completely sure how to set travis up, I assume you checked manually that the tests are running :)

PR #11, opened 2020-04-12
Create covid_graphs python package.

### LINKED ISSUE #4: Create setup.py
Python utilities should be installable and easily runnable from anywhere.

miskosz:
Creates a python package:

- Rename `graphs` -> `covid_graphs` to have a less generic name.
- All the dependencies in `setup.py` (for now).
- Created a submodule structure:
-- `pb` for protobuf generated code
-- `scripts` for command line tools (see also note below)

Issues:
- **Testing:** Locally I tested just `fit_atg_model` (it has tests) and managed to run `server.py`. Might have broken something else.
- **Docker:** I fought and lost, can't make `docker-compose` run locally. What do we use the `covid_graphs/Dockerfile` for? It seems it can't do much anyway since it does not have access to `data/`.
- **Command line scripts:** I think it is non-standard to have mixed non-executable and executable logic in a file. Typically a library provides a function and then a command line script (that uses `argparse` or its more modern alternative `click`) just invokes the library. I think this is not too difficult of a task for a next PR. I have to understand though how are those scripts used first.


covid_graphs/Dockerfile
```diff
@@ -6,6 +6,5 @@ RUN apt update \
     && apt install -y python3-pip python3-numpy python3-pandas \
     && rm -rf /tmp/* /var/lib/apt/lists/* /root/.cache/*
 
-COPY requirements.txt .
-
-RUN pip3 install -r requirements.txt
+COPY . .
+RUN pip3 install .
```
miskosz: I did not manage to make this work locally. @lukipuki how is this file used?

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -25,13 +28,14 @@
 covid19_normal_app = CountryReport.create_dashboard(args.data_dir, server, GraphType.Normal)
 covid19_semilog_app = CountryReport.create_dashboard(args.data_dir, server, GraphType.SemiLog)
 covid19_loglog_app = CountryReport.create_dashboard(args.data_dir, server, GraphType.LogLog)
-covid19_heatmap_app = HeatMap(args.simulated_polynomial).create_app(server)
-covid19_heatmap_exponential_app = HeatMap(args.simulated_exponential).create_app(server)
+# covid19_heatmap_app = HeatMap(args.simulated_polynomial).create_app(server)
+# covid19_heatmap_exponential_app = HeatMap(args.simulated_exponential).create_app(server)
```
lukipuki: These were forgotten I guess. I can send you the protobufs, so that you don't have to comment these.
```suggestion
covid19_heatmap_app = HeatMap(args.simulated_polynomial).create_app(server)
covid19_heatmap_exponential_app = HeatMap(args.simulated_exponential).create_app(server)
```

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -1,9 +1,12 @@
 #!/usr/bin/env python3
-from country import CountryReport, GraphType
-from heat_map import HeatMap
 from flask import Flask, render_template, redirect, url_for
 import argparse
 import glob
+from pathlib import Path
```
lukipuki, line 5: Are you using `Path` anywhere?

covid_graphs/Dockerfile
lukipuki: `COPY . .` looks wrong (it should be a no-op).
Strange that pip3 install doesn't work. I'll try it myself.

docker-compose.yml
```diff
@@ -2,13 +2,13 @@ version: "2.2"
 
 services:
   dash_server:
-    build: graphs/
+    build: covid_graphs/
     image: accu_dash
     volumes:
-      - "./graphs:/covid19/graphs:ro"
+      - "./covid_graphs:/covid19/covid_graphs:ro"
```
lukipuki, line 8: Assuming we get pip install work inside Dockerfile ...
```suggestion
```

docker-compose.yml
```diff
@@ -2,13 +2,13 @@ version: "2.2"
...
       - "./data:/covid19/data:ro"
     working_dir: "/covid19"
     ports:
       - 127.0.0.1:8081:8080
     restart: always
-    command: "/covid19/graphs/server.py data/ data/results-poly.pb data/results-exp.pb"
+    command: "/covid19/covid_graphs/covid_graphs/scripts/server.py data/ data/results-poly.pb data/results-exp.pb"
```
lukipuki, line 14:
```suggestion
    command: "server.py data/ data/results-poly.pb data/results-exp.pb"
```

covid_graphs/Dockerfile
lukipuki: The following works
```
FROM ubuntu:19.10

ENV DEBIAN_FRONTEND=noninteractive
RUN echo 'APT::Get::Assume-Yes "true";' >> /etc/apt/apt.conf
RUN apt update \
    && apt install -y python3-pip python3-numpy python3-pandas \
    && rm -rf /tmp/* /var/lib/apt/lists/* /root/.cache/*

COPY . /covid19_graphs
RUN pip3 install /covid19_graphs
RUN rm -rf /covid19_graphs
```

covid_graphs/setup.py
```diff
@@ -0,0 +1,26 @@
...
+        "pandas",
+        "protobuf",
+        "plotly",
+        "pytest",
+        "scipy",
+
+    ],
+    dependency_links=[],
+    include_package_data=True,
+    python_requires=">=3.7",
```
lukipuki, line 21: Why do we need 3.7?

covid_graphs/setup.py
```diff
@@ -0,0 +1,26 @@
...
+    packages=find_packages(),
```
lukipuki:
```suggestion
    packages=find_namespace_packages(),
```
And you can then drop all `__init__.py` empty files

covid_graphs/covid_graphs/scripts/converter.py
```diff
@@ -5,9 +5,7 @@
 from datetime import datetime, timedelta
 from google.protobuf import text_format
 
-sys.path.append('../graphs')
```
lukipuki: This file can be deleted completely, it served a purpose once ...

covid_graphs/covid_graphs/scatter_plot.py
```diff
@@ -5,9 +5,10 @@
 from google.protobuf import text_format
```
lukipuki: scatter_plot.py should be also in scripts

lukipuki:
README.md links to the old `graphs/README.md`, should be renamed too.

covid_graphs/setup.py
lukipuki: Using Python 3.7 will be problematic. Travis runs Ubuntu 18.04 and it only has Python 3.6.

lukipuki: I thought it would be a problem with Travis, but they support it: https://docs.travis-ci.com/user/reference/bionic/#python-support

covid_graphs/covid_graphs/scripts/server.py
miskosz: Thanks, missed that :)

covid_graphs/setup.py
miskosz: We definitely want dataclasses and typing for variables. I think 3.7 is the earliest that supports that.

covid_graphs/setup.py
miskosz: This is magic, had no idea it exists :)

covid_graphs/covid_graphs/scripts/server.py
miskosz: Not anymore, this slipped...

covid_graphs/covid_graphs/scatter_plot.py
miskosz: I kept it here since the it seems the graph creation logic is nontrivial. I can move it for now to scripts, but probably we should split the functionality from the command line call... unless it is not used anywhere else.

lukipuki: OK, I'll refactor it next time I use it.

covid_graphs/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Running without Docker
+Alternatively, you can run the server locally without Docker. Create a virtual
+environment, install the `covid_graphs` package an run the server.
+```sh
+python3.7 -m venv your/path/to/venv
+source your/path/to/venv/bin/activate
+pip install -e .
+server.py data build/results-poly.pb build/results-exp.pb
+```
```
miskosz, line 20: Expanded on this section.

covid_graphs/covid_graphs/scripts/server.py
miskosz: And back again :)

docker-compose.yml
miskosz: I was not able to make `index.html` load by running the global version of `server.py` script. I suggest to keep for now.

covid_graphs/README.md
```diff
@@ -0,0 +1,31 @@
+# Graphs and visualizations related to COVID-19
+
+## Running the whole server
+
+In the root directory of the repository, run `docker-compose`.
+```sh
+docker-compose up --build # Optionally add -d for deamon
+```
+You can then access the server locally at `http://127.0.0.1:8081`.
```
lukipuki:
```suggestion
You can then access the server locally at [localhost:8081](http://127.0.0.1:8081)`.
```

covid_graphs/README.md
lukipuki: This part will use the port 8080, which is confusing, because it's different from `docker-compose`. I suggest changing the port inside server.py to 8081 as well (you'll need to modify docker-compose as well).

lukipuki:
:+1:

lukipuki (approved):

docker-compose.yml
lukipuki: You need too keep this mount if we aren't running server.py using pip install.

PR #12, opened 2020-04-13
Refactor stats

### LINKED ISSUE #5: Do not use constexpr for log_factorial

lukipuki:
I've moved a lot of functionality inside `stats.h` and `person.h`, as `simulation.cpp` should be really only about setting everything up and let the important logic live elsewhere.

### DIFF
--- a/src/generators.h
+++ b/src/generators.h
@@ -18,14 +18,17 @@ class GeneratorInterface {
     return deltas;
   }
 
-  virtual auto CreateDeltas(uint32_t t0, uint32_t count) const -> std::vector<double> = 0;
+  [[nodiscard]] virtual auto CreateDeltas(uint32_t t0, uint32_t count) const
+      -> std::vector<double> = 0;
   virtual ~GeneratorInterface() = default;
 };
 
 class ExponentialGenerator : public GeneratorInterface {
  public:
   ExponentialGenerator(double gamma1, double gamma2) : gamma1_(gamma1), gamma2_(gamma2) {}
-  auto CreateDeltas(uint32_t t0, uint32_t count) const -> std::vector<double> override {
+  // Creates deltas as defined in COR01.pdf.
+  [[nodiscard]] auto CreateDeltas(uint32_t t0, uint32_t count) const
+      -> std::vector<double> override {
     t0 = std::min(t0, count - 1);
     std::vector<double> result = ExponentialPrefix(gamma1_, t0);
     while (result.size() < count) {
@@ -44,7 +47,9 @@ class PolynomialGenerator : public GeneratorInterface {
  public:
   PolynomialGenerator(double gamma1, double polynomial_exponent)
       : gamma1_(gamma1), polynomial_exponent_(polynomial_exponent) {}
-  auto CreateDeltas(uint32_t t0, uint32_t count) const -> std::vector<double> override {
+  // Creates deltas with polynomial growth after t0.
+  [[nodiscard]] auto CreateDeltas(uint32_t t0, uint32_t count) const
+      -> std::vector<double> override {
     t0 = std::min(t0, count - 1);
     std::vector<double> values;
     for (int i = 1; i <= count + 1; ++i) {
--- a/src/stats.h
+++ b/src/stats.h
@@ -14,8 +14,8 @@ constexpr std::array<double, kDecadesCount> kPopulationAge = {0.11, 0.10, 0.12, 
                                                               0.13, 0.13, 0.07, 0.03};
 constexpr std::array<double, kDecadesCount> kDeathProbabilities = {
     0.002, 0.002, 0.002, 0.002, 0.004, 0.013, 0.036, 0.08, 0.148};
+constexpr double kDeathThreshold = 0.5;
 constexpr uint32_t kSymptomsLength = 28;
-
 constexpr uint32_t kFactorialLength = 5000;
 
 // Class containing all the logic for generating people. Each person:
--- a/src/stats.h
+++ b/src/stats.h
@@ -83,7 +83,8 @@ class Stats {
 
   // Generates the number of infected based on deltas. See Rado Harman's COR01.pdf for the
   // definition of delta_t.
-  auto GenerateInfected(const std::vector<double>& deltas, uint32_t count) -> std::vector<uint32_t> {
+  auto GenerateInfected(const std::vector<double>& deltas, uint32_t count)
+      -> std::vector<uint32_t> {
     assert(count <= deltas.size());
     std::vector<uint32_t> infected;
     for (uint32_t i = 0; i < count; ++i) {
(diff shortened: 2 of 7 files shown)


src/generators.h
```diff
@@ -18,14 +18,17 @@ class GeneratorInterface {
     return deltas;
   }
 
-  virtual auto CreateDeltas(uint32_t t0, uint32_t count) const -> std::vector<double> = 0;
+  [[nodiscard]] virtual auto CreateDeltas(uint32_t t0, uint32_t count) const
+      -> std::vector<double> = 0;
```
miskosz, line 22: My C++ knowledge is way behind to understand this :)

src/stats.h
```diff
@@ -12,64 +12,109 @@ constexpr uint32_t kPopulationSize = 5450000;
...
+    return x / (x + y);
+  }
+
+  // Quantile function of beta distribution B(1, b).
+  static auto qbeta(double b, double quantile) -> double { return 1 - pow(1 - quantile, 1.0 / b); }
+
+  std::vector<double> log_factorials_;
+  // For each decade of life 'i', calculate 'bs_[i]', such that
+  // 'Pr[B(1, bs_[i]) > 0.5] = kDeathProbabilities[i]'.
+  std::array<double, kDecadesCount> bs_;
```
miskosz: What does `bs_` stand for?

src/stats.h
```diff
@@ -12,64 +12,109 @@ constexpr uint32_t kPopulationSize = 5450000;
...
+
+  // Logarithm of the distance function. More details in Rado Harman's COR01.pdf.
+  auto LogDistance(uint32_t z, uint32_t c) -> double {
+    if (z + c == 0) {
+      return 0;
+    }
+    assert(z >= 0 && c >= 0);
+    uint32_t max = std::max(z, c) + 1;
+    for (uint32_t i = log_factorials_.size(); i < max; ++i) {
+      log_factorials_.push_back(log_factorials_.back() + std::log(i));
```
miskosz: Minor: Maybe in the future you can have `LogFactorials` as a simple class that manages `log_factorials_` internally, so that users don't have to do that themselves.

src/stats.h
```diff
@@ -12,64 +12,109 @@ constexpr uint32_t kPopulationSize = 5450000;
...
+  }
 
-auto generate_age(std::mt19937* generator) -> uint32_t {
-  std::uniform_real_distribution<> dis(0.0, 1.0);
-  double generated = dis(*generator);
-  double sum = 0;
-  uint32_t res = 0;
-  for (; sum + kPopulationAge[res] <= generated; ++res) {
-    sum += kPopulationAge[res];
+  // Poisson distribution, only used for testing.
```
miskosz: It is used also in `GenerateInfected`, is the comment correct?

src/stats.h
```diff
@@ -12,64 +12,109 @@ constexpr uint32_t kPopulationSize = 5450000;
...
-// Logarithm of the probability used in the distance function.
-auto log_distance_probability(uint32_t z, uint32_t c) -> double {
-  assert(z + c > 0);
-  assert(z < kFactorialLength);
-  assert(c < kFactorialLength);
-  double positive = (z + c) * std::log((z + c) / 2.);
-  double negative = z + c + log_factorials[z] + log_factorials[c];
-  return positive - negative;
-}
+  std::random_device rd_;
```
miskosz: Is `rd_` needed? It seems it is used just for `random_generator_` initialisation.

src/stats.h
```diff
@@ -12,64 +12,109 @@ constexpr uint32_t kPopulationSize = 5450000;
...
-auto beta_distribution(double alpha, double beta, std::mt19937* gen) -> double {
-  std::gamma_distribution<> X(alpha, 1), Y(beta, 1);
-  double x = X(*gen), y = Y(*gen);
-  return x / (x + y);
-}
+// Class containing all the logic for generating people. Each person:
+// * is generated according to population age distribution
+// * has symptoms' severity generated based on their age
+// * has disease length generated based on symptoms severity
+class Stats {
```
miskosz: Minor: Maybe it should be called `PopulationDistribution` or `PopulationModel` or something similar.

miskosz (approved):
I went through it. I have not worked recently with C++ production code (can't check style) and did not study the simulation model (can't check logic), but I went at least through `stats.h`. (I skimmed through the tests very lightly.) LGTM though :)

src/stats.h
lukipuki: `bs_[i]` is b^(i) from the PDF: Screenshot_20200413_180416

I've added the 's' to signal plural (that's the typical convention for arrays, but I get that it's confusing in this case. I can change it to `b_[i]` instead.

src/stats.h
lukipuki: Right, `PopulationModel` sounds good. The name of the files `stats.h` was more fitting earlier.

src/generators.h
lukipuki: - virtual - same as in other languages
- `auto function() -> return_type` is the same as `def function(): -> return_type` in Python
- `= 0` means a true abstract function that has to be implemented by classes that are derived from this one.

src/stats.h
lukipuki: Yeah, though there's some speed overhead with classes. Probably doesn't matter much here though.

src/stats.h
lukipuki: You are right, I wrote it earlier when this comment was actually true and GenerateInfected was in another file.

src/stats.h
lukipuki: Changed to `b_[i]`.

src/stats.h
lukipuki: Dropped it.

src/generators.h
miskosz: And `[[nodiscard]]`? :)

src/stats.h
miskosz: Both `b_` and `bs_` are fine with me, I was just wondering if there is a more descriptive name. It is ok if that's how the term is called.

src/generators.h
Aj0SK: Yeah, I also wondered why is [[nodiscard]] used :)

src/stats.h
Aj0SK: Even with optimizations turned on?

src/generators.h
lukipuki: That's just so that you don't forget to use the return value. https://en.cppreference.com/w/cpp/language/attributes/nodiscard

src/generators.h
```diff
@@ -18,14 +18,17 @@ class GeneratorInterface {
...
 };
 
 class ExponentialGenerator : public GeneratorInterface {
  public:
   ExponentialGenerator(double gamma1, double gamma2) : gamma1_(gamma1), gamma2_(gamma2) {}
-  auto CreateDeltas(uint32_t t0, uint32_t count) const -> std::vector<double> override {
+  // Creates deltas as defined in COR01.pdf.
+  [[nodiscard]] auto CreateDeltas(uint32_t t0, uint32_t count) const
+      -> std::vector<double> override {
     t0 = std::min(t0, count - 1);
```
Aj0SK, line 32: I have question regarding this line. Is count always bigger than 0?

lukipuki: Yeah, It's the length in days, which is now at least 40 days (since March 2, when our data starts).

src/stats.h
lukipuki: Yeah, inlining gets tricky, compiler has its own mind and rules :)

PR #13, opened 2020-04-13
Refactor how server is run.

miskosz:
* Use `click` instead of `argparse`.
* Create a function to run the server and expose `covid_graphs.run_server` command.
* Fix distributing `index.html` with the package. Simlify `docker-compose.yml`.

### DIFF
--- a/covid_graphs/README.md
+++ b/covid_graphs/README.md
@@ -16,8 +16,9 @@ environment, install the `covid_graphs` package an run the server.
 python3.7 -m venv your/path/to/venv
 source your/path/to/venv/bin/activate
 pip install -e . # With -e the package will automatically reload with any local changes.
-server.py data build/results-poly.pb build/results-exp.pb
+covid_graphs.run_server --data-dir data/ --simulated-polynomial data/results-poly.pb --simulated-exponential data/results-exp.pb
 ```
+Run `covid_graphs.run_server --help` for help on command parameters.
 
 
 ## Running standalone graphs
--- a/covid_graphs/setup.py
+++ b/covid_graphs/setup.py
@@ -7,6 +7,8 @@ setup(
     version="1.0",
     description="COVID-19 graph plotting",
     install_requires=[
+        "click",
+        "click_pathlib",
         "dash",
         "Flask",
         "numpy",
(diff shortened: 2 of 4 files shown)


covid_graphs/setup.py
```diff
@@ -17,9 +19,13 @@
...
     packages=find_namespace_packages(),
-    scripts=["covid_graphs/scripts/prepare_data.py", "covid_graphs/scripts/scatter_plot.py", "covid_graphs/scripts/server.py"],
-    package_data={"covid_graphs": ["covid_graphs/scripts/index.html"]},
+    scripts=["covid_graphs/scripts/prepare_data.py", "covid_graphs/scripts/scatter_plot.py"],
+    entry_points={
+        "console_scripts": [
+            "covid_graphs.run_server=covid_graphs.scripts.server:run_server"
+        ]
+    },
+    package_data={"": ["*.html"]},
```
miskosz, line 30: This line was the tricky part. The syntax is
```
package_data={package_name: [glob_pattern_to_include_1, ...]}
```
It does not work though when I specify e.g. `{"covid_graphs": ["*.html"]}`, I think it might be because we use namespace packages (no `__init__.py` files).

covid_graphs/setup.py
```diff
@@ -17,9 +19,13 @@
...
     dependency_links=[],
-    include_package_data=True,
     python_requires=">=3.7",
     packages=find_namespace_packages(),
-    scripts=["covid_graphs/scripts/prepare_data.py", "covid_graphs/scripts/scatter_plot.py", "covid_graphs/scripts/server.py"],
-    package_data={"covid_graphs": ["covid_graphs/scripts/index.html"]},
+    scripts=["covid_graphs/scripts/prepare_data.py", "covid_graphs/scripts/scatter_plot.py"],
+    entry_points={
+        "console_scripts": [
+            "covid_graphs.run_server=covid_graphs.scripts.server:run_server"
```
lukipuki, line 27: I like entry_points, I've started converting scatter_plot to it.

covid_graphs/README.md
```diff
@@ -16,8 +16,9 @@ environment, install the `covid_graphs` package an run the server.
 python3.7 -m venv your/path/to/venv
 source your/path/to/venv/bin/activate
 pip install -e . # With -e the package will automatically reload with any local changes.
-server.py data build/results-poly.pb build/results-exp.pb
+covid_graphs.run_server --data-dir data/ --simulated-polynomial data/results-poly.pb --simulated-exponential data/results-exp.pb
```
lukipuki, line 19: Not related to this review: I thought about this line and we have it somewhat inconsistent.

- We give it a directory, which contains data for 7 countries. It tries to intelligently pick all country files and skips those that aren't present.
- Then we give it as arguments two files in the same data directory, but with the full path, the app does no guessing and no skipping.

I think eventually we should drop the `--simulated-polynomial` and `--simulated-exponential` arguments and just use `--data-dir`.

lukipuki (approved):

PR #14, opened 2020-04-13
More entry points

### LINKED ISSUE #4: Create setup.py

lukipuki:
I've migrated all graphs into entry points.

The scripts could use some refactoring for better future reuse, but we'll do it some other time.


covid_graphs/covid_graphs/prepare_data.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python3
```
miskosz: These lines can be removed in all files that have been converted.

covid_graphs/setup.py
```diff
@@ -21,10 +21,13 @@
...
     python_requires=">=3.7",
     packages=find_namespace_packages(),
-    scripts=["covid_graphs/scripts/prepare_data.py", "covid_graphs/scripts/scatter_plot.py"],
     entry_points={
         "console_scripts": [
-            "covid_graphs.run_server=covid_graphs.scripts.server:run_server"
+            "covid_graphs.run_server = covid_graphs.scripts.server:run_server",
+            "covid_graphs.country = covid_graphs.country:main",
+            "covid_graphs.heat_map = covid_graphs.heat_map:main",
+            "covid_graphs.scatter_plot = covid_graphs.scatter_plot:main",
```
miskosz: I'd suggest commands that are a bit more verbose, e.g. sth like:
```
covid_graphs.create_country_plot
covid_graphs.create_heat_map
covid_graphs.create_scatter_plot
```
(Maybe `show` instead of `create`.)

miskosz (approved):
Nice :)

covid_graphs/covid_graphs/prepare_data.py
lukipuki: Done

covid_graphs/setup.py
lukipuki: In principle I agree, but these are command line tools and typing all of them is a hassle, plus they take too much space on the command line. I suggest shortening the prefix, as it doesn't carry that much information. For example:
```suggestion
            "cg.scatter_plot = covid_graphs.scatter_plot:main",
```

miskosz: I disagree, I think clarity trumps convenience here. It is actually not bad at all - the command autocompletes on tabs, and usually at least I type the command once and reuse over again.

miskosz: Also it is a suggestion, feel free to go with what you think is best.

miskosz: Also a small note, for `run_server` I enabled short versions of the params, `-d`, `-p` and `-e`.

lukipuki: My autocomplete doesn't work and I actually have a pretty good setup. When I type the command once, I get auto-suggestions though:
Screenshot_20200414_132212

I'll think about it for a few more hours, but I think I'll go with the shortened prefix.

lukipuki: After some thought I went with your 'show*' suggestions.

PR #17, opened 2020-04-14
Can't install python packages with old version of setuptools

michalburger1:
If you follow the manual python installation instructions, you may get the following error:

```
ImportError: cannot import name 'find_namespace_packages' from 'setuptools'
```

This is fixed by updating the setuptools package.


lukipuki:
Please use setuptools of version at least 40.1.0: https://setuptools.readthedocs.io/en/latest/history.html#v40-1-0

lukipuki (approved):

PR #18, opened 2020-04-14
New predictions

### LINKED ISSUE #15: Country graphs should support different starting times of curves
We've got new predictions which have different starting times than the old ones and it'd be interesting to be able to compare both. Same goes for testing different curve fitting algorithms, you might need to plot curves with different starting times and visually compare them against each other.

### LINKED ISSUE #16: Create a dashboard for the new PLparameters.pdf

lukipuki:


covid_graphs/covid_graphs/country_report.py
```diff
@@ -0,0 +1,24 @@
...
+class CountryReport:
+    def __init__(self, data_dir, country_tuple):
+        with open(os.path.join(data_dir, f'{country_tuple.name}.data'), "rb") as f:
+            country_data = CountryData()
+            text_format.Parse(f.read(), country_data)
+            self.name = country_data.name
+            positive = np.array([day.positive for day in country_data.stats])
+            self.dead = np.array([day.dead for day in country_data.stats])
+            self.recovered = np.array([day.recovered for day in country_data.stats])
+            self.active = positive - self.recovered - self.dead
```
miskosz: I see now why I was getting negative numbers, the meaning is something like "newly active per day". So if more people get recovered/die than get infected we get a negative number. Would you mind renaming to `newly_active`?

I see we compute `cumulative_active` below. Which of the two notions is predicted by AtgModel?

covid_graphs/covid_graphs/formula.py
```diff
@@ -0,0 +1,64 @@
+import math
+from datetime import datetime, timedelta
+import numpy as np
+from collections import namedtuple
+from enum import Enum
+
+EXPONENT = 6.23
+
+# TODO(lukas): Move to dataclass?
```
miskosz, line 9: I think this can be done later. In the current form, making a class out of ATG_formula would be an option. If we want to support other formulas then of course it would be more complex.

covid_graphs/covid_graphs/formula.py
```diff
@@ -0,0 +1,64 @@
...
+                   min_case_count)
+
+
+class EvaluatedFormula():
+    def __init__(self,
+                 formula,
+                 cumulative_active,
+                 first_idx,
+                 last_idx,
+                 first_date, xaxis_type):
```
miskosz: nit: format

covid_graphs/covid_graphs/formula.py
```diff
@@ -0,0 +1,64 @@
...
+        self.text = formula.text
+        # First index: t=1
+        start_idx = np.argmax(cumulative_active >= formula.min_case_count)
```
miskosz, line 42: This is a little cryptic :) But I guess its fine.

covid_graphs/covid_graphs/formula.py
```diff
@@ -0,0 +1,64 @@
...
+        length = last_idx - start_idx
+        self.y = formula.lambd(np.arange(length) + 1)
+        self.date_list = [(first_date + timedelta(days=d)).strftime('%Y-%m-%d')
+                          for d in range(start_idx, last_idx + 1)]
```
miskosz, line 46: For now it is fine. I think though that we can somehow refactor the formula concept such that it thas besides A and TG also a parameter T0. Then this shifts can be implemented in a more elegant way.

miskosz: Is `last_idx` supposed to be inclusive? Note that `len(self.date_list) = length + 1`.

miskosz (approved):
LGTM. I must say it gets slightly difficult to read, using types and passing named parameters could help :) I think the functionality can be surely written in a nicer way, but as it is it gets the job done, so let's go with it.

covid_graphs/covid_graphs/country_report.py
lukipuki: I think we have the convention to use `daily_(infected|active|dead)` elsewhere, e.g. in some of the protos. Would prefer using that.

AtgModel computes cumulative numbers.

covid_graphs/covid_graphs/formula.py
lukipuki: I think I had it consistent at some point but then changed it back to how it looks in our current production. It doesn't matter much, whether we round up or down around the inflection point. At least I'll make it consistent now.

lukipuki: Done

covid_graphs/covid_graphs/country_report.py
lukipuki: Done

PR #19, opened 2020-04-14
[WIP] Fit test

michalburger1:
Do not merge.

Just a quick demo of how the data fitting works. Apply and watch.

### DIFF
--- a/covid_graphs/covid_graphs/country.py
+++ b/covid_graphs/covid_graphs/country.py
@@ -11,6 +11,7 @@ import numpy as np
 import os
 from google.protobuf import text_format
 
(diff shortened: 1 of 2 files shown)


michalburger1:
Not sure whether the results are useful or not. You can disable the x0 argument if you want more conservative results. The predictions vary wildly and I'm concerned that they're not very meaningful.

miskosz (changes requested):
So that we don't merge.

lukipuki:
GitHub has a "Compare" button, please use it in the future. Screenshot_20200417_165148

Closing.

PR #20, opened 2020-04-14
Use pathlib with Country graph.

miskosz:
Polishing the code before making bigger changes.

### DIFF
--- a/covid_graphs/covid_graphs/country_report.py
+++ b/covid_graphs/covid_graphs/country_report.py
@@ -18,8 +18,8 @@ class CountryReport:
     def __init__(self, data_dir: Path, country: Country):
         """Constructs a numpy representation of data read from 'data_dir' a given country."""
         country_data = CountryData()
-        data_file = Path(data_dir / f'{country.name}.data')
-        text_format.Parse(data_file.read_text(), country_data)
+        data_file_path = Path(data_dir / f'{country.name}.data')
+        text_format.Parse(data_file_path.read_text(), country_data)
 
         self.name = country.name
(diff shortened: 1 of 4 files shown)


covid_graphs/covid_graphs/country_report.py
```diff
@@ -1,24 +1,36 @@
...
-            self.daily_recovered = np.array([day.recovered for day in country_data.stats])
-            self.daily_active = daily_positive - self.daily_recovered - self.daily_dead
-            self.date_list = [
-                f"{day.date.year}-{day.date.month:02d}-{day.date.day:02d}"
-                for day in country_data.stats
-            ]
+    def __init__(self, data_dir: Path, country: Country):
+        """Constructs a numpy representation of data read from 'data_dir' a given country."""
+        country_data = CountryData()
+        data_file = Path(data_dir / f'{country.name}.data')
```
lukipuki: Doesn't the file need to be closed?

lukipuki (approved):

covid_graphs/covid_graphs/country_report.py
miskosz: This does not open the file, just creates a path object (essentially just a smarter string). Will rename the variable for clarity.

miskosz: For completeness, from https://docs.python.org/3/library/pathlib.html:
```
Path.read_text(encoding=None, errors=None)
...
The file is opened and then closed.
```

lukipuki: Smart!

PR #21, opened 2020-04-14
Add predictions database.

miskosz:
Separate predictions from the rest of the code. The functionality should be unchanged from before.

This is a step towards serving graphs in a more modular way.

I will add a test before merging, but please have a look already.

### DIFF
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -11,12 +11,12 @@ import numpy as np
 from pathlib import Path
 from typing import List
 
-from . import predictions
-from .predictions import CountryPrediction
+from .predictions import CountryPrediction, prediction_db
 from .country_report import CountryReport
 from .formula import ATG_formula, EvaluatedFormula, Formula, XAxisType
 
 
+# TODO: Remove after graph display refactoring.
 PREDICTION_DATE = '2020-03-29'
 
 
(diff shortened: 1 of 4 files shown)


covid_graphs/covid_graphs/country_graph.py
```diff
@@ -9,10 +9,14 @@
...
+from typing import List
 
-from .country_report import Country, CountryReport
+from . import predictions
+from .predictions import CountryPrediction
+from .country_report import CountryReport
 from .formula import ATG_formula, EvaluatedFormula, Formula, XAxisType
 
+
 PREDICTION_DATE = '2020-03-29'
```
lukipuki, line 20: This constant needs to go away, but I'm fine doing that in another PR. Please add a TODO.
```suggestion
```

covid_graphs/covid_graphs/predictions.py
```diff
@@ -0,0 +1,101 @@
...
+@dataclass
+class CountryPrediction:
+    prediction_event: PredictionEvent
+    country: str
+    formula: Formula
+
+
+BK_20200329 = PredictionEvent(name="bk_20200329", date=datetime.date(2020, 3, 29))
+BK_20200413 = PredictionEvent(name="bk_20200413", date=datetime.date(2020, 4, 13))
+OTHER = PredictionEvent(name="other", date=datetime.date(2020, 4, 13)) # TODO: When did we add SVK prediction?
```
lukipuki: This is not a prediction, but rather just trying to fit a polynomial on Slovak data . I think you can remove it from here and we should rather have it in the "fitting" script, when he have one.

lukipuki (approved):
Awesome refactoring, types really help readability.

I ran this myself and noticed that the order of countries is different between reloads of the server. The order should be stable and (preferably) similar to the order of countries on our original site.
We could also write the prediction date into the curve title, so that it's visible in the legend.

covid_graphs/covid_graphs/predictions.py
miskosz: It is your prediction :)

miskosz:
Ready to merge. I decided to skip the tests and added a TODO.

I did two changes since yesterday:

- Make a class for serving predictions
- Remove dependency in `CountryReport` on the list of predictions.

**Feel free to merge yourself if there are no further comments.**

covid_graphs/covid_graphs/country_report.py
```diff
@@ -4,24 +4,17 @@
...
-    def __init__(self, data_dir: Path, country: Country):
+    def __init__(self, data_dir: Path, country_name: str):
         """Constructs a numpy representation of data read from 'data_dir' a given country."""
         country_data = CountryData()
-        data_file_path = Path(data_dir / f'{country.name}.data')
+        data_file_path = Path(data_dir / f'{country_name}.data')
         text_format.Parse(data_file_path.read_text(), country_data)
 
-        self.name = country.name
+        self.name = country_name
```
lukipuki: Just a note here: these are not the same names.

- country_name is used for data storage, so we have `UK.data` or `USA.data`.
- country.name is longer, we store `United Kingdom` there, for example.

So I'd prefer keeping it the old way, and renaming `country_name` to `short_country_name`.

miskosz: Nice catch!

`Country` should probably became a class as well (storing names/ids for various purposes). Leaving for later PRs.

PR #23, opened 2020-04-15
EvaluatedFormula is a curve

### LINKED ISSUE #16: Create a dashboard for the new PLparameters.pdf

lukipuki:
Episode #8301 of Naming Is Hard

### DIFF
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -1,4 +1,3 @@
-from collections import namedtuple
 from enum import Enum
 from plotly.graph_objs import Figure, Layout, Scatter
 import click
@@ -9,12 +8,10 @@ import dash_html_components as html
 from flask import Flask
 import numpy as np
 from pathlib import Path
-from typing import List
 
-from .predictions import CountryPrediction, prediction_db
+from .predictions import prediction_db
 from .country_report import CountryReport
-from .formula import ATG_formula, EvaluatedFormula, Formula, XAxisType
-
+from .formula import Curve, XAxisType
 
 # TODO: Remove after graph display refactoring.
 PREDICTION_DATE = '2020-03-29'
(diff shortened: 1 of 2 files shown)


covid_graphs/covid_graphs/formula.py
```diff
@@ -31,8 +31,8 @@ def ATG_formula(TG, A, exponent, min_case_count=200):
                    min_case_count)
 
 
-class EvaluatedFormula():
-    """Class containing an evaluated formula.
+class Curve():
```
miskosz, line 34: Naming is hard indeed :D

miskosz: `Trace` would be an option.

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -219,5 +217,7 @@ def create_dashboard(data_dir: Path, server: Flask, graph_type: GraphType=GraphT
     type=str,
 )
 def show_country_plot(data_dir: Path, country_name: str):
-    country_graph = CountryGraph(data_dir=data_dir, country_name=country_name, graph_type=GraphType.Normal)
+    country_graph = CountryGraph(data_dir=data_dir,
+                                 country_name=country_name,
+                                 graph_type=GraphType.Normal)
```
miskosz, line 222: Btw. I plan to add to CI a style checker over the weekend, together with type checker and a testing framework.

miskosz (approved):
LGTM!

Food for thought (not this PR): From my experience, classes are used in python sparsely - only when the object needs to persist some state. In general, pure functions are preferred, placing them in the right submodule (=file) manages the logical grouping. This helps keeping the code modular and also easy to test.

In the case of `Curve`, I would instead create a dataclass with attributes `t, y, date_list` etc., and even add `first_idx` and `last_idx`. Then `create_curves` can be a pure function that returns a `Curve`.

covid_graphs/covid_graphs/country_graph.py
miskosz: Actually I meant automatic code formatter :)

lukipuki:
Yeah `create_curves` is a static method, so the class is really just a simple data holder. I'm not familiar with dataclasses yet (one step at a time!).

PR #26, opened 2020-04-15
Adding all the new predictions

### LINKED ISSUE #16: Create a dashboard for the new PLparameters.pdf

lukipuki:


covid_graphs/covid_graphs/formula.py
```diff
@@ -31,19 +31,29 @@ def ATG_formula(TG, A, exponent, min_case_count=200):
...
+    It's the first day for which the number of cumulative active is more than
+    formula.min_case_count.
+    """
+    return np.argmax(cumulative_active > formula.min_case_count)
+
+
 class Curve():
-    """Class containing an curve derived from a formula.
+    """
+    Class containing an curve derived from a formula.
```
miskosz: Nit: a curve

covid_graphs/covid_graphs/formula.py
```diff
@@ -31,19 +31,29 @@ def ATG_formula(TG, A, exponent, min_case_count=200):
...
 
 
+def first_day_of_curve(cumulative_active, formula):
+    """
+    Returns the first day of the curve formed by 'formula'.
+
+    It's the first day for which the number of cumulative active is more than
+    formula.min_case_count.
+    """
+    return np.argmax(cumulative_active > formula.min_case_count)
```
miskosz, line 41: Did you intentionally change `>` to `>=`?

data/download_all.zsh
```diff
@@ -1,17 +1,22 @@
...
 short_names["United Kingdom"]=UK
+short_names["New Zealand"]=NZ
 
 python3.7 -m venv --system-site-packages venv && source venv/bin/activate
 pip install $1/
-for country in Germany Iran Italy Spain "United Kingdom" US
+
+countries=("United Kingdom" Korea Austria Iceland Jordan Switzerland Croatia Australia Canada
+  Germany Israel Italy Malaysia "New Zealand" Spain Belgium Chile Czechia Latvia
+  Lithuania Netherlands Norway Portugal US)
```
miskosz: You are using `US` here and not `United States`, not sure if intentional.

lukipuki:
FYI, there's something incorrect about the predictions - I found inconsistencies between us and Katka/Rišo for Croatia and Czechia. They are investigating. Until then, we shouldn't merge this.

miskosz (approved):
LGTM

miskosz:
@lukipuki I'd suggest commenting Croatia and Czechia out and merging like that.

We are not displaying the graphs anyway, are we?

covid_graphs/covid_graphs/formula.py
lukipuki: Yes, that's how Katka does the fitting. I spotted in on Lithuania, where the difference is 12 days :smile:

miskosz:
Oh, those two were just examples, weren't they :D

data/download_all.zsh
lukipuki: Thanks for noticing, that's a bug.

lukipuki:
Yes, those are just examples, I suspect a bigger issue somewhere (could be on either side).

But you are right, let's merge and we'll wait for their comments.

covid_graphs/covid_graphs/formula.py

PR #28, opened 2020-04-16
Introduce SimulationReport data class

lukipuki:
The old SimulationReport became SimulationGraph

### DIFF
--- a/covid_graphs/covid_graphs/scatter_plot.py
+++ b/covid_graphs/covid_graphs/scatter_plot.py
@@ -9,7 +9,7 @@ import click_pathlib
 import numpy as np
 
 from .country_report import CountryReport
-from .simulation_report import create_simulation_reports
+from .simulation_report import create_simulation_reports, GrowthType
 
 EXTENSION = 10
 
@@ -22,16 +22,20 @@ class GraphType(Enum):
         return self.value
 
 
-def create_dates(list_of_lists: List[List[int]], date_list: List[str]) -> List[str]:
-    day_indexes = chain.from_iterable(range(len(l)) for l in list_of_lists)
-    return [date_list[index] for index in day_indexes]
+def create_dates(simulation_results: List[List[int]], date_list: List[str]) -> List[str]:
+    """
+    Generates a flat list of dates for a list of simulation results.
+
+    Typically used as x-axis labels of the simulation results.
+    """
+    return [date_list[index] for l in simulation_results for index in range(len(l))]
 
 
-class SimulationGraph():
+class SimulationGraph:
     def __init__(self, country_data_file: Path, simulation_pb2_file: Path):
         reports = create_simulation_reports(simulation_pb2_file)
 
-        self.best_error, self.best_b0, self.best_gamma2 = 1e20, None, None
+        self.best_error, self.best_b0, self.best_gamma2 = float("inf"), None, None
         for report in reports:
             if report.error > self.best_error:
                 continue
@@ -41,10 +45,11 @@ class SimulationGraph():
             self.deltas = report.deltas
             self.best_b0 = report.b0
             self.prefix_length = report.prefix_length
-            self.param_name = report.param_name
             self.best_param = report.param
             self.best_error = report.error
+            self.growth_type = report.growth_type
 
+        self.param_name = "alpha" if self.growth_type == GrowthType.Polynomial else "gamma2"
         print(f"b_0={self.best_b0}, {self.param_name}={self.best_param} is the best fit "
               f"for prefix_length={self.prefix_length}, error={self.best_error}")
 
--- a/covid_graphs/covid_graphs/scatter_plot.py
+++ b/covid_graphs/covid_graphs/scatter_plot.py
@@ -1,15 +1,15 @@
 from datetime import timedelta
 from enum import Enum
 from itertools import accumulate, chain
-import numpy as np
 from pathlib import Path
 from plotly.graph_objs import Figure, Layout, Scatter
+from typing import List
 import click
 import click_pathlib
+import numpy as np
 
 from .country_report import CountryReport
-from .pb.simulation_results_pb2 import SimulationResults
-
+from .simulation_report import create_simulation_reports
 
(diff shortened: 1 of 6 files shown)


covid_graphs/covid_graphs/simulation_report.py
```diff
@@ -0,0 +1,42 @@
...
+class SimulationReport():
+    daily_positive: List[List[int]]
+    daily_infected: List[List[int]]
+    deltas: List[float]
+    b0: int
+    prefix_length: int
+    error: float
+    # We want to store either alpha or gamma2. We could have two classes, one having alpha and the
+    # other gamma2 (in C++ I would make a template). Don't know what's best practice in these
+    # situations in Python.
```
miskosz, line 27: I am not sure what is the exact use case, my hunch is that something fishy is going on and there is a more fitting abstraction. Nevertheless, I see two options.

A) You can use inheritance:
```
class SimulationReportAlpha(SimulationReport):
    ...
```
I'm not completely sure how it is compatible with dataclasses though; the decorator adds some convenience methods (most notably an initialiser), so there's that.

B)
You can use:
```
alpha: Optional[float]
gamma2: Optional[float]
```
You need `from typing import Optional`. Also your code needs to manage to use only one of them.

covid_graphs/covid_graphs/simulation_report.py
```diff
@@ -0,0 +1,42 @@
...
+    param_name: str
+    param: float
```
miskosz, line 28: I'd consider adding `growth_type: GrowthType` and move the enum definition to this file.

covid_graphs/covid_graphs/scatter_plot.py
```diff
@@ -22,42 +22,36 @@ def __str__(self):
         return self.value
 
 
-class SimulationReport():
+def create_dates(list_of_lists: List[List[int]], date_list: List[str]) -> List[str]:
```
miskosz: Can you add a function comment? It is difficult to know what this does.

covid_graphs/covid_graphs/scatter_plot.py
```diff
@@ -22,42 +22,36 @@ def __str__(self):
...
+    day_indexes = chain.from_iterable(range(len(l)) for l in list_of_lists)
+    return [date_list[index] for index in day_indexes]
```
miskosz: Without knowing the intended functionality, it seems this could be rewritten as:
```
return [
    date_list[index]
    for l in list_of_lists
    for index in range(len(l))
]
```

covid_graphs/covid_graphs/scatter_plot.py
```diff
@@ -22,42 +22,36 @@ def __str__(self):
...
+
+
+class SimulationGraph():
     def __init__(self, country_data_file: Path, simulation_pb2_file: Path):
-        simulation_results = SimulationResults()
-        simulation_results.ParseFromString(simulation_pb2_file.read_bytes())
+        reports = create_simulation_reports(simulation_pb2_file)
 
         self.best_error, self.best_b0, self.best_gamma2 = 1e20, None, None
```
miskosz: Nit: You can also do this:
```
self.best_error, self.best_b0, self.best_gamma2 = float("inf"), None, None
```

miskosz (approved):
LGTM

covid_graphs/covid_graphs/simulation_report.py
lukipuki: We have `oneof` in the proto and I'd like to simulate the same behavior: https://github.com/lukipuki/COVID-19-simulation/blob/master/src/simulation_results.proto#L6-L9

covid_graphs/covid_graphs/scatter_plot.py
lukipuki: Done

covid_graphs/covid_graphs/scatter_plot.py
lukipuki: Done

covid_graphs/covid_graphs/scatter_plot.py
lukipuki: Done

covid_graphs/covid_graphs/simulation_report.py
miskosz: I think "What does `oneof` in the proto try to achieve?" is the right question to ask.

If I understand correctly, it seems there are two types of simulation types. Each type has a different set of parameters, in the future maybe there will be one with two float parameters, who knows.

If that's the case, you could define the two simulation types separately:
```
@dataclass
class PolynomialGrowth:
    alpha: float

@dataclass
class ExponentialGrowth:
    gamma2: float
```
and then have
```
class SimulationReport:
    ...
    simulation_params: Union[PolynomialGrowth, ExponentialGrowth]
```

Bottomline: I think that would be an overkill, I'd keep what you have if we are just printing the params. Or B) above if we are using them in python code :)

covid_graphs/covid_graphs/simulation_report.py
```diff
@@ -0,0 +1,42 @@
+from dataclasses import dataclass
+from pathlib import Path
+from typing import List
+
+from .pb.simulation_results_pb2 import SimulationResults
+
+
+@dataclass
+class SimulationReport():
```
miskosz: I just noticed, I think in python3 you can drop the parentheses (I'm like 88% sure).

lukipuki: Done

covid_graphs/covid_graphs/simulation_report.py
lukipuki: Done

covid_graphs/covid_graphs/simulation_report.py
lukipuki: Since I've added growth_type, we don't *really* need inheritance or anything, so I'll keep it as it if for now. At some point I might add inheritance or optional, but it's very low priority right now.

PR #29, opened 2020-04-16
Create separate dashboards for different prediction events.

miskosz:
What has changed?

- Moved webpage creation out of `country_graph.py`
- A dashboard is created per *prediction event* and per *graph type*
- Added dashboard links to the index page

What needs to be done in future PRs:

- The text at the top of a graph page does not describe what is going on.

### DIFF
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -10,9 +10,6 @@ from .predictions import prediction_db, CountryPrediction
 from .country_report import CountryReport
 from .formula import Curve, XAxisType
 
-# TODO: Remove after graph display refactoring.
-PREDICTION_DATE = '2020-03-29'
-
 
 class GraphType(Enum):
     Normal = 'normal'
@@ -36,6 +33,10 @@ class CountryGraph:
         # TODO(miskosz): We assume there is only one country.
         # This might change soon though if we want to have a country dropdown in one graph.
         country_name = country_predictions[0].country
+
+        # TODO(lukas): Figure out better strategy for more predictions
+        if len(country_predictions) >= 1:
+            self.prediction_date = country_predictions[0].prediction_event.date.strftime("%Y-%m-%d")
 
         # Due to plotly limitations, we can only have graphs with dates on the x-axis when we
         # aren't using logs.
@@ -72,7 +73,7 @@ class CountryGraph:
                  line=dict(width=2, dash='dot')) for curve in self.curves
         ]
         try:
-            prediction_date = self.date_list.index(PREDICTION_DATE)
+            prediction_date = self.date_list.index(self.prediction_date)
             # Add green zone marking the data available at the prediction date.
             shapes.append(
                 dict(type="rect",
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -2,14 +2,11 @@ from enum import Enum
 from plotly.graph_objs import Figure, Layout, Scatter
 import click
 import click_pathlib
-import dash
-import dash_core_components as dcc
-import dash_html_components as html
-from flask import Flask
 import numpy as np
 from pathlib import Path
+from typing import List
 
-from .predictions import prediction_db
+from .predictions import prediction_db, CountryPrediction
 from .country_report import CountryReport
 from .formula import Curve, XAxisType
 
@@ -28,13 +25,24 @@ class GraphType(Enum):
 
 class CountryGraph:
     """Constructs a graph for a given country"""
-    def __init__(self, data_dir: Path, country_name: str, graph_type: GraphType = GraphType.Normal):
-        country_predictions = prediction_db.predictions_for_country(country=country_name)
+    def __init__(
+        self,
+        data_dir: Path,
+        country_predictions: List[CountryPrediction],
+        graph_type: GraphType = GraphType.Normal,
+    ):
         self.graph_type = graph_type
+
+        # TODO(miskosz): We assume there is only one country.
+        # This might change soon though if we want to have a country dropdown in one graph.
+        country_name = country_predictions[0].country
+
         # Due to plotly limitations, we can only have graphs with dates on the x-axis when we
         # aren't using logs.
         axis_type = XAxisType.Dated if graph_type == GraphType.Normal else XAxisType.Numbered
+
         report = CountryReport(country_data_file=data_dir / f'{country_name}.data')
+
         first_idx, last_idx, self.curves = Curve.create_curves(
             [prediction.formula for prediction in country_predictions],
             report.cumulative_active,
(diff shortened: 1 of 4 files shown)


covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -0,0 +1,102 @@
...
+            ### Notes about the graphs
+
+            * Dashed lines are the predictions, solid red lines are the real active
+              cases. Black dotted lines mark the predicted maximums.
+            * We've now added new prediction made on 2020-04-13 by the same authors. There's a
+              second dashed line for Italy, Spain, USA and Germany.
+            * France has been excluded, since they [screwed up daily data reporting]({france_link}).
+            """,
+                     dangerously_allow_html=True)
+    ]
```
lukipuki: Aký používaš editor? Lebo ti tuším uchádzajú konce riadkov v súboroch.

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -36,31 +42,53 @@ def run_server(data_dir: Path, simulated_polynomial: Path, simulated_exponential
...
-    covid19_heatmap_app = HeatMap(str(simulated_polynomial)).create_app(server)
-    covid19_heatmap_exponential_app = HeatMap(str(simulated_exponential)).create_app(server)
+    _create_prediction_apps(server=server, data_dir=data_dir)
+    _create_simulation_apps(server, simulated_polynomial, simulated_exponential)
 
     @server.route("/")
     def home():
         return render_template("index.html")
 
     @server.route("/covid19/")
```
lukipuki: This is the URL that we pasted all over the internet, we should redirect that one.

```suggestion
    @server.route("/covid19/normal")
```

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -36,31 +42,53 @@ def run_server(data_dir: Path, simulated_polynomial: Path, simulated_exponential
...
+        "semilog": GraphType.SemiLog,
+        "loglog": GraphType.LogLog,
+    }
+    route_pairs = itertools.product(event_by_route.keys(), graph_type_by_route.keys())
 
-    @server.route("/covid19/semilog")
-    def covid19_semilog():
-        return covid19_semilog_app.index()
+    prediction_apps = {
+        (event_route, graph_type_route): country_dashboard.create_dashboard(
```
lukipuki, line 73: What sourcery is this? A lambda without the `lambda` keyword?

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -36,31 +42,53 @@ def run_server(data_dir: Path, simulated_polynomial: Path, simulated_exponential
...
-        return covid19_loglog_app.index()
+    @server.route("/covid19/predictions/<event>/<graph_type>")
+    def covid19_predictions(event: str, graph_type: str):
+        return prediction_apps[(event, graph_type)].index()
+
+
+def _create_simulation_apps(
+    server: Flask, simulated_polynomial: Path, simulated_exponential: Path
+):
+    covid19_heatmap_app = HeatMap(str(simulated_polynomial)).create_app(server)
```
lukipuki, line 90: I've fixed to take path
```suggestion
    covid19_heatmap_app = HeatMap(simulated_polynomial).create_app(server)
```

lukipuki: I've fixed it to take `Path`
```suggestion
    covid19_heatmap_app = HeatMap(simulated_polynomial).create_app(server)
```

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -36,31 +42,53 @@ def run_server(data_dir: Path, simulated_polynomial: Path, simulated_exponential
...
+            prediction_event=event_by_route[event_route],
+            graph_type=graph_type_by_route[graph_type_route],
+        )
+        for event_route, graph_type_route in route_pairs
+    }
 
-    @server.route("/covid19/loglog")
-    def covid19_loglog():
-        return covid19_loglog_app.index()
+    @server.route("/covid19/predictions/<event>/<graph_type>")
```
lukipuki, line 82: Is this some magic Flask templated routing?

lukipuki (approved):

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -2,14 +2,11 @@
...
-import dash_html_components as html
-from flask import Flask
 import numpy as np
 from pathlib import Path
+from typing import List
 
-from .predictions import prediction_db
+from .predictions import prediction_db, CountryPrediction
 from .country_report import CountryReport
 from .formula import Curve, XAxisType
```
lukipuki, line 12: The constant 2 lines below (GitHub doesn't allow comments there) `PREDICTION_DATE` shouldn't be a constant but rather depend on the prediction event.

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -0,0 +1,102 @@
...
+        name=f'COVID-19 {graph_type}',
+        url_base_pathname=f'/covid19/{prediction_event.name}/{graph_type}/',
+        server=server,
+        external_scripts=[
+            'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/MathJax.js?config=TeX-MML-AM_CHTML'
+        ],
+    )
+    app.title = 'COVID-19 predictions'
+    if graph_type != GraphType.Normal:
+        app.title += f' on a {graph_type} graph'
```
lukipuki, line 40: We should add similar logic to the `html.H1` element - the main headline should reflect the graph type

covid_graphs/covid_graphs/scripts/country_dashboard.py
miskosz: Sublime. When I add a linter it is going to catch these I hope :)

covid_graphs/covid_graphs/scripts/server.py
miskosz: Yep, good idea. Slightly unrelated, why do we route everything through `/covid19/` prefix?

covid_graphs/covid_graphs/scripts/server.py
miskosz: LOL :D No, it is a dictionary comprehension:
```py
dct = {key: val for x in whatever}
```

miskosz: `prediction_apps` is of type `Dict[Tuple[str, str], Dash]`. I create the dash apps before registering routes with flask, this is where I store them. It is not super elegant. I think it might be not necessary though, did not investigate.

covid_graphs/covid_graphs/scripts/server.py
miskosz: Yes, this is flask routing... I have read the docs :) This is actually quite a standard thing when you think about it, e.g. when you have `lukipuki.blog.com/article/841083`.

covid_graphs/covid_graphs/country_graph.py
miskosz: Agreed. I did not change it yet since it needs changing the graph layout logic.

covid_graphs/covid_graphs/scripts/server.py
lukipuki: The prefix is there, since the full URL is `https://graphs.<domain>/covid19`. If we take it away, COVID-19 would not be present in the URL, which I think is wrong. You might argue that the prefix should be added by Nginx instead - I'd agree with it, but that's extra work. Or we can change the subdomain to `covid-graphs.<domain>`, but I think we've missed that train :smile:

This is yet another episode of Naming is Hard.

covid_graphs/covid_graphs/scripts/server.py
lukipuki: Oh, dictionary comprehension, I forgot that exists!

covid_graphs/covid_graphs/scripts/server.py
lukipuki: Yeah, I thought about using it exactly for normal/semilog/loglog before, but skipped it because of other priorities.

PR #30, opened 2020-04-17
Restart on data change

### LINKED ISSUE #25: Reload data when it's updated. Do not restart the server when source code is updated.
We wanted to react to data updates, so we run the Flask server like so:
```
server.run(host="0.0.0.0", port=8081, debug=True, extra_files=data_dir.glob("*.data"))
```

However, this restarts the server even when code is updated, which is not what we want.

Aj0SK:


covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -70,4 +73,15 @@ def covid19_heatmap_polynomial():
...
 
-    server.run(host="0.0.0.0", port=8081, debug=True, extra_files=data_dir.glob("*.data"))
+    i = inotify.adapters.InotifyTree(str(data_dir))
+
+    p = Process(target=server.run, kwargs=dict(host="0.0.0.0", port=8081))
+    p.start()
+    while (True):
+        events = list(i.event_gen(yield_nones=False, timeout_s=1))
+        if len(events) != 0:
+            p.terminate()
```
lukipuki: Add a print statement, that you are restarting the server based on file changes inside the data directory.

lukipuki (approved):

PR #31, opened 2020-04-17
Simplify server's parameters

### LINKED ISSUE #25: Reload data when it's updated. Do not restart the server when source code is updated.

### LINKED ISSUE #27: Allow starting server even without simulation protos
Server should start even without simulation data. It should show an error page instead.

lukipuki:
The simulation protos are part of the data directory. If they don't
exist, we show an error page.

I've disabled the restarting functionality, since it's broken.

### DIFF
--- a/covid_graphs/covid_graphs/heat_map.py
+++ b/covid_graphs/covid_graphs/heat_map.py
@@ -67,11 +67,18 @@ def create_heat_map(simulation_table: SimulationTable):
     return figure
 
 
-def create_heat_map_dashboard(simulation_pb2_file: Path, server: Flask):
+def create_heat_map_dashboard(simulation_pb2_file: Path, growth_type: GrowthType, server: Flask):
+    app = dash.Dash(
+        name=f'COVID-19 {growth_type} heat map',
+        server=server,
+        url_base_pathname=f"/covid19/heatmap/{growth_type}/",
+        external_scripts=[
+            'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/MathJax.js?config=TeX-MML-AM_CHTML'
+        ])
     try:
         simulation_reports = create_simulation_reports(simulation_pb2_file)
-    except ValueError:
-        app = dash.Dash(name=f'COVID-19 heat map', server=server)
+    except (ValueError, FileNotFoundError) as e:
+        print(e)
         app.layout = html.Div(dcc.Markdown(f"""
             # Error in parsing simulation proto file
 
@@ -82,13 +89,6 @@ def create_heat_map_dashboard(simulation_pb2_file: Path, server: Flask):
         return app
 
     growth_type = next(iter(simulation_reports)).growth_type
-    app = dash.Dash(
-        name=f'COVID-19 {growth_type} heat map',
-        server=server,
-        url_base_pathname=f"/covid19/heatmap/{growth_type}/",
-        external_scripts=[
-            'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/MathJax.js?config=TeX-MML-AM_CHTML'
-        ])
     app.title = 'Heat map of a COVID-19 stochastic model, with {growth_type} growth'
 
     simulation_tables = group_data(simulation_reports)
@@ -121,5 +121,7 @@ def create_heat_map_dashboard(simulation_pb2_file: Path, server: Flask):
     type=click_pathlib.Path(exists=True),
 )
 def show_heat_map(simulation_protofile):
-    app = create_heat_map_dashboard(simulation_protofile, True)
+    # TODO(lukas): This can be exponential growth, but if the file fails to parse there's no way of
+    # knowing it.
+    app = create_heat_map_dashboard(simulation_protofile, GrowthType.Polynomial, True)
     app.run_server(host="0.0.0.0", port=8080)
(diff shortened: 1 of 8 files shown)


covid_graphs/covid_graphs/heat_map.py
```diff
@@ -13,7 +13,14 @@
 
 class HeatMap():
     def __init__(self, simulation_pb2_file: Path):
-        reports = create_simulation_reports(simulation_pb2_file)
+        try:
+            reports = create_simulation_reports(simulation_pb2_file)
+            self.parsed_reports = True
+        except:
```
miskosz: I would discourage using a `try` block here as it can hide other exceptions. Non-existence of a file is now a valid state, and can be easily check programatically:
```
if not simulation_pb2_file.is_file():
   ...
   return
```

I also think that from the design point of view this is not the best place for the check. One other possible way to achieve the same would be to move the check to the caller, e.g. along these lines:

- Decouple graph creation from dash dashboard creation
- `HeatMap` initialiser accepts `SimulationResults` proto class
- Dashboard creation reads the proto and instantiates `HeatMap`. If it can't then serves an error page instead of the dashboard.

This way `HeatMap` does not have an ad hoc way of signalling that something went wrong.

I'm not sure if you want to spend effort figuring it out, I'll approve if you don't care that much.

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -56,20 +36,27 @@ def home():
...
         if len(events) != 0:
             print("Data changed. Restarting server...")
             p.terminate()
             p.join()
-            p = Process(target=server.run, kwargs=dict(host="0.0.0.0", port=8081))
+            p = Process(target=_run_flask_server, kwargs=dict(server=server, data_dir=data_dir))
             p.start()
             print("Server restarted.")
-        sleep(60)
+        sleep(10)
```
miskosz, line 54: So we watch for changes in the server code? I thought we'd have an external watch script that would kill and restart the python server process.

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -56,20 +36,27 @@ def home():
...
+
+
+def _run_flask_server(server: Flask, data_dir: Path):
+    _create_prediction_apps(server=server, data_dir=data_dir)
+    _create_simulation_apps(server=server, data_dir=data_dir)
```
miskosz, line 59: It seems inconsistent that these two have moved but `home()` and `covid19_redirect()` have stayed.

covid_graphs/covid_graphs/heat_map.py
miskosz: I have realised you don't really need approval to merge 😄

lukipuki: This code is correct with respect to the current implementation, as `create_simulation_reports` can only fail when the proto is missing or fails to parse.

I agree that that's not the best practice, since code eventually changes. However, I think the right place to parse the proto is inside `simulation_report.py`. Right now no other source file knows about the existence of the `SimulationResults` proto and I think that's great! The only thing that anyone else should know about the proto file is it's path.

What I could agree with is initializing HeatMap with `SimulationReport`, but that means that `create_simulation_reports` might fail (as it's the case now).

I'm really tired now, so I'll think about the right course of action tomorrow..

covid_graphs/covid_graphs/scripts/server.py
lukipuki: TL;DR: Seems like too much price to pay for a bit more consistency.

The routes don't have to be updated, AFAIK. To be consistent, we would have to do a move one way or the other. Moving e.g. `home()` is worse perfomance-wise (and readability as well).

The opposite move would require some refactoring. And actually fixing the restart wasn't even my goal with this PR, I just noticed that the fix from https://github.com/lukipuki/COVID-19-simulation/issues/25 wasn't working and was affecting my PR too. Someone else can make this nicer if they want.

covid_graphs/covid_graphs/scripts/server.py
lukipuki: Initially when we used YAML and it took 2 minutes to reload the server, we had plans to reload partial things like individual graphs for countries that have changed.

I think the external script just adds in more complications and Python can do the same job.

covid_graphs/covid_graphs/heat_map.py
lukipuki: This ended up super-complicated (see commit eebd11d), since HeatMap grew to be a big spaghetti mess. This is the diff of my last commit in `server.py`:
```
-    covid19_heatmap_polynomial_app = HeatMap(simulated_polynomial).create_app(server)
-    covid19_heatmap_exponential_app = HeatMap(simulated_exponential).create_app(server)
+    covid19_heatmap_polynomial_app = create_heat_map_dashboard(simulated_polynomial, server)
+    covid19_heatmap_exponential_app = create_heat_map_dashboard(simulated_exponential, server)
```

I've added a dataclass and a few associated functions. Makes for better error reporting. `create_heat_map_dashboard` might fail if the proto fails to parse.

miskosz: You are right, parsing the proto inside `simulation_report.py` makes sense :) I did not really think through the "perfect" error catching logic, and in the end I think what you had was good enough.

PR #32, opened 2020-04-17
Improved graphics

lukipuki:
- solid curves
- dashed line only goes to the maximum and has the same color as the curve
- green zone is less opaque
- the blue background is gone
- vertical grid lines are gone

Screenshot_20200417_230634

### DIFF
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -23,10 +23,10 @@ class GraphType(Enum):
 class CountryGraph:
     """Constructs a graph for a given country"""
     def __init__(
-        self,
-        data_dir: Path,
-        country_predictions: List[CountryPrediction],
-        graph_type: GraphType = GraphType.Normal,
+            self,
+            data_dir: Path,
+            country_predictions: List[CountryPrediction],
+            graph_type: GraphType = GraphType.Normal,
     ):
         self.graph_type = graph_type
 
@@ -62,10 +62,8 @@ class CountryGraph:
             self.t = np.arange(len(self.cumulative_active)) + 1
 
     def create_country_figure(self):
-        colors = ['rgb(31, 119, 180)', '#bcbd22', 'violet'][:len(self.curves)]
+        colors = ['SteelBlue', 'Purple', 'Green'][:len(self.curves)]
 
-        # max_y = max(curve.maximal_y for curve in self.curves)
-        # bottom_y_margin = 0.08 * max_y
         shapes = [
             # Add vertical dotted lines marking the maxima
             dict(type="line",
@@ -73,7 +71,7 @@ class CountryGraph:
                  y0=0,
                  x1=curve.t[curve.maximal_idx],
                  y1=curve.maximal_y,
-                 line=dict(width=2, dash='dash', color=color))
+                 line=dict(width=2, dash="dash", color=color))
             for color, curve in zip(colors, self.curves)
         ]
         try:
@@ -87,7 +85,7 @@ class CountryGraph:
                      y0=0,
                      y1=1,
                      fillcolor="LightGreen",
-                     opacity=0.5,
+                     opacity=0.3,
                      layer="below",
                      line_width=0))
         except ValueError:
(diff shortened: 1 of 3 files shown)


covid_graphs/covid_graphs/country_graph.py
```diff
@@ -23,10 +23,10 @@ def __str__(self):
...
     """Constructs a graph for a given country"""
     def __init__(
-        self,
-        data_dir: Path,
-        country_predictions: List[CountryPrediction],
-        graph_type: GraphType = GraphType.Normal,
+            self,
+            data_dir: Path,
+            country_predictions: List[CountryPrediction],
+            graph_type: GraphType = GraphType.Normal,
```
miskosz, line 29: Something fishy happened here :)

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -83,28 +85,35 @@ def create_country_figure(self):
...
+        layout = Layout(
+            title=f"Active cases in {self.name}",
+            xaxis=dict(
+                autorange=True,
+                title=f'Day [starting at the {self.min_case_count}th case]',
+                showgrid=False,
+            ),
+            yaxis=dict(
+                autorange=True,
+                tickformat='.0f',
```
miskosz: Interesting, does this print e.g. "80k" as in your screenshot? From the docs I thought it should be `"~s"`: https://github.com/d3/d3-format/blob/master/README.md#locale_format
(Also I like the "80k" format.)

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -66,9 +66,8 @@ def _get_header_content(prediction_event: PredictionEvent, title: str):
             'On 2020-03-30, mathematicians Katarína Boďová and Richard Kollár ',
             html.A('made predictions about 7 countries', href=prediction_link),
             f'. The data available up to that point (until {prediction_date_str}) is in the ',
-            html.Span('green zone', style={'color': 'green'}),
-            f'. Data coming after {prediction_date_str} is in the ',
-            html.Span('blue zone.', style=dict(color='blue')),
+            html.Span('light green zone', style={'color': 'lightgreen'}),
```
miskosz: Just wondering if `background-color` attribute would be more fitting.

miskosz (approved):
❤️

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -62,15 +62,17 @@ def __init__(
...
 
     def create_country_figure(self):
+        colors = ['SteelBlue', 'Purple', 'Green'][:len(self.curves)]
+
         shapes = [
             # Add vertical dotted lines marking the maxima
             dict(type="line",
-                 yref="paper",
                  x0=curve.t[curve.maximal_idx],
                  y0=0,
```
miskosz, line 71: Nemali by sme dat y0>0? (Napr. `y0=1`.) Mam pocit ze semilog a loglog grafy vyzeraju divne...

covid_graphs/covid_graphs/country_graph.py
lukipuki: That's formatting according to `yapf`

covid_graphs/covid_graphs/country_graph.py
lukipuki: 80k is confusing to general public.

I'd prefer something along
```
d3.format(",.2r")(4223);  // grouped thousands with two significant digits, "4,200"
```

It has a divider (a comma) and does some rounding. Though I'd skip the rounding.

covid_graphs/covid_graphs/scripts/country_dashboard.py
lukipuki: Will try it out tomorrow.

covid_graphs/covid_graphs/country_graph.py
lukipuki: Yes, but the `yaxis` range should be also reduced to `[1, max_value]`. Without it it looks even weirder.

PR #34, opened 2020-04-18
Python testing scripts.

miskosz:
Add python testing.

I'm not that confident with bash, comments are welcome.

Also I added the `__init__.py` files, I think it is the simplest and most reliable solution.

Note: Still setting up Travis.

### DIFF
--- a/covid_graphs/README.md
+++ b/covid_graphs/README.md
@@ -24,11 +24,9 @@ Run `covid_graphs.run_server --help` for help on command parameters.
 ## Local development
 
 This script creates a virtual environment with all development dependencies.
-Don't forget to activate the environment:
 ```sh
 ./create_dev_venv.sh
-source .venv/bin/activate
-```
+source .venv/bin/activate # Don't forget to activate the environment
 
 To run tests, simply run:
 ```sh
--- a/covid_graphs/README.md
+++ b/covid_graphs/README.md
@@ -15,7 +15,6 @@ environment, install the `covid_graphs` package an run the server.
 ```sh
 python3.7 -m venv your/path/to/venv
 source your/path/to/venv/bin/activate
-pip install --upgrade setuptools
 pip install -e . # With -e the package will automatically reload with any local changes.
 covid_graphs.run_server --data-dir data
 ```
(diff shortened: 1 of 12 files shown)


lukipuki:
I think for Travis we can skip virtualenv and just use pip. The virtual machine gets thrown away anyway. What would be the added value of using virtualenv there? (The only one I know of setting up $PATH which we don't use)

miskosz:
It is useful to run tests in local development.
(Also in general it is nicer, say e.g. if conan had incompatible setup.)

lukipuki:
I'm not disputing local development, just that we can skip setting up virtualenv in Travis.

miskosz:
Ready for review.

I thought I'd reuse the venv in CI, in the end it is useful just for developing locally. Also we duplicate the package installation now in 3 places:
* dockerfile
* travis
* create venv

I think it might be difficult to consolidate, so I'd just keep it as is for now.

covid_graphs/README.md
```diff
@@ -11,21 +11,31 @@ You can then access the server locally at [localhost:8081](http://127.0.0.1:8081
...
-## Running standalone graphs
+## Local development
 
-For quick development or data examination, running standalone graphs is useful.
+This script creates a virtual environment with all development dependencies.
+Don't forget to activate the environment:
+```sh
+./create_dev_venv.sh
+source .venv/bin/activate
+```
```
lukipuki: I think this order is more clear.
```suggestion
```sh
./create_dev_venv.sh
source .venv/bin/activate # Don't forget to activate the environment
```
```

File: covid_graphs/create_dev_venv.sh [hunk 2]
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+set -euo pipefail
```
lukipuki, line 2: I think this is the same as
```suggestion
#!/usr/bin/env bash -eu
set -o pipefail
```

I've been using `bash -e` and I'm not familiar with `-u`. Perhaps even `-o pipefail` can be put there, e.g. `!/usr/bin/env bash -euo pipefail`

lukipuki (approved):

covid_graphs/create_dev_venv.sh
miskosz: I took inspiration from our codebase :) This is what "the internets" say: https://vaneyckt.io/posts/safer_bash_scripts_with_set_euxo_pipefail/

miskosz: I pipefail can't go there, so I've kept it as it was.

PR #37, opened 2020-04-18
More precise predictions

lukipuki:
I've also reformatted code using 'black'


miskosz (approved):
LGTM.

Also glad you managed to make black work :)

PR #39, opened 2020-04-19
Update HTML header.

### LINKED ISSUE #35: Update text on dashboards
We've now added second dashboard with predictions from April 12, but its text needs to be updated.

-  The formula is now N(t) = (A/TG) ⋅ (t/TG)^alpha / e^{t/TG} (previously alpha=6.23)
-  We have changed the blue background color, made dashed -> solid, dotted -> dashed. All of these have to be updated in both dashboards.
-  The following sentence should be removed (doesn't hold on the new dashboards) "We've now added new prediction made on 2020-04-13 by the same authors. There's a second dashed line for Italy, Spain, USA and Germany."
-  France has been excluded from March 29 predictions, but we should also point out that Iran's data is very likely flawed too. We can readd Iran back to the dashboard, but perhaps provide some links. I don't know what to link though. I usually trust comments at slatestarcodex.com, so here's few links, more are in the article itself:
https://slatestarcodex.com/2020/04/10/coronalinks-4-10-second-derivative/#comment-878300).

miskosz:
Screenshot 2020-04-19 at 18 17 25

I'm still not sure what to do about France and Iran.

Crossref: #35

### DIFF
--- a/covid_graphs/covid_graphs/scripts/country_dashboard.py
+++ b/covid_graphs/covid_graphs/scripts/country_dashboard.py
@@ -37,8 +37,13 @@ def create_dashboard(
     )
     app.title = 'COVID-19 predictions of Boďová and Kollár'
 
-    content = _get_header_content(prediction_event, app.title)
-    content += [html.Hr()]
+    content = _get_header_content(app.title)
+
+    next_day = prediction_event.date + timedelta(days=1)
+    content += [
+        html.Hr(),
+        html.H1(f"{next_day.strftime('%B %d')} predictions")
+    ]
     buttons = [
         dcc.Dropdown(id='country-short-name',
                      options=[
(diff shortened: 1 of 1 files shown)


covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -77,53 +82,64 @@ def update_graph(graph_type_str, country_short_name):
...
+
+            * <em>N</em>(<em>t</em>) = (<em>A</em>/<em>T</em><sub><em>G</em></sub>) ⋅
+              (<em>t</em>/<em>T</em><sub><em>G</em></sub>)<sup>α</sup> /
+              e<sup><em>t</em>/<em>T</em><sub><em>G</em></sub></sup>
+
+            Where:
+
+            * <em>t</em> is time in days counted from a country-specific "day one"
+            * <em>N(t)</em> is the number of active cases (cumulative positively tested minus recovered and deceased)
+            * <em>A</em>, <em>T<sub>G</sub></em> and <em>α</em> are country-specific parameters
```
lukipuki: I've left the formula as HTML, since I got it for free from pandoc from my TeX formula. However, in the text, we can use markdown's stars to mark italics.
```suggestion
            * *t* is time in days counted from a country-specific "day one"
            * *N(t)* the number of active cases (cumulative positively tested minus recovered and deceased)
            * *A*, *T<sub>G</sub>* and *α* are country-specific parameters
```

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -77,53 +82,64 @@ def update_graph(graph_type_str, country_short_name):
...
+                html.Li("Dashed line marks the culmination of the prediction"),
+                html.Li("Red line is observed number of active cases"),
+                html.Li(
+                    children=[
+                        "Data available until the date of prediction is in ",
+                        html.Span('light green zone', style={'background-color': 'lightgreen'}),
+                    ]
+                ),
+            ]
+        )
```
lukipuki, line 144: I think this would read better as Markdown (you can include HTML for the span using `dangerously_allow_html=True`.

lukipuki (approved):
Given that there is one header that works for both prediction sets, what do you think about adding another dropdown for April 11 and March 30? We would only have one page with predictions.

covid_graphs/covid_graphs/scripts/country_dashboard.py
miskosz: I tried. It did not work for me, no idea why.

miskosz:
Yes, I think it would be logical to have only one page.

PR #40, opened 2020-04-19
Put predictions on one page

lukipuki:
Screenshot_20200419_211828

### DIFF
--- a/covid_graphs/covid_graphs/scripts/country_dashboard.py
+++ b/covid_graphs/covid_graphs/scripts/country_dashboard.py
@@ -15,11 +15,11 @@ from covid_graphs.predictions import prediction_db, PredictionEvent, BK_20200411
 def create_graphs(
     reports: List[CountryReport], prediction_event: PredictionEvent,
 ):
-    report_from_name = {report.short_name: report for report in reports}
+    report_by_name = {report.short_name: report for report in reports}
 
     # Note: We silently assume there is only one prediction per country.
     country_graphs = [
-        CountryGraph(report_from_name[country_prediction.country], [country_prediction])
+        CountryGraph(report_by_name[country_prediction.country], [country_prediction])
         for country_prediction in prediction_db.predictions_for_event(prediction_event)
     ]
     country_graphs.sort(key=lambda graph: graph.short_name)
@@ -33,7 +33,7 @@ def create_dashboard(
     print("Creating dashboard for prediction graphs.")
     prediction_events = prediction_db.get_prediction_events()
     prediction_events.sort(key=lambda event: event.date, reverse=True)
-    prediction_event_from_name = {
+    prediction_event_by_name = {
         prediction_event.name: prediction_event for prediction_event in prediction_events
     }
 
@@ -100,7 +100,7 @@ def create_dashboard(
             dict(label=graph.long_name, value=graph.short_name)
             for graph in graph_dict[prediction_event_name]
         ]
-        next_day = prediction_event_from_name[prediction_event_name].date + timedelta(days=1)
+        next_day = prediction_event_by_name[prediction_event_name].date + timedelta(days=1)
         return options, f"{next_day.strftime('%B %d')} predictions"
 
     @app.callback(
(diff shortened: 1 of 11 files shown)


covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -5,90 +5,136 @@
...
-        server: Flask,
-        prediction_event: PredictionEvent,
+def create_graphs(
+    reports: List[CountryReport], prediction_event: PredictionEvent,
 ):
-    # TODO(miskosz): Don't use print.
-    print(f"Creating dashboard for {prediction_event.name} graphs.")
-
-    predictions = prediction_db.predictions_for_event(prediction_event)
+    report_from_name = {report.short_name: report for report in reports}
```
miskosz: Minor side note: My personal preference is "by" in place of "from". Either is fine though.

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -5,90 +5,136 @@
...
-                       options=[{
-                           'label': graph_type.value,
-                           'value': graph_type.name,
-                       } for graph_type in [GraphType.Normal, GraphType.SemiLog, GraphType.LogLog]],
-                       value='Normal',
-                       labelStyle={'display': 'inline-block'})
+        dcc.Dropdown(
+            id="prediction-event",
+            options=[
+                dict(label=(event.date + timedelta(days=1)).strftime("%B %d"), value=event.name)
```
miskosz, line 68: I think we should expand the `PredictionEvent` class. There should be two date fields: `publish_date`/`prediction_date` and `data_until_date`. (Maybe with better names.)

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -5,90 +5,136 @@
...
+                for event in prediction_events
+                if event != OTHER
+            ],
+            value=BK_20200411.name,
+            style={"width": "40%"},
```
miskosz, line 73: We should add a css for the whole page to make it nicer btw :)

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -53,30 +52,17 @@ def covid19_redirect():
...
-        event_route:
-        country_dashboard.create_dashboard(data_dir=data_dir,
-                                           server=server,
-                                           prediction_event=event_by_route[event_route])
-        for event_route in event_by_route
-    }
-
-    # TODO(lukas): add redirect @server.route("/covid19/predictions/latest")
-    @server.route("/covid19/predictions/<event>")
+    @server.route("/covid19/predictions/")
```
miskosz, line 63: I noticed we don't specify the trailing `/` for other routes, just checking that it is fine.

miskosz (approved):
Nicely done!

covid_graphs/covid_graphs/scripts/country_dashboard.py
lukipuki: Fixed.

I used to have 'by' there actually. But then when I said it out loud in my head it didn't sound good, so I changed it.

I almost always use 'by' in code in these situations though. Just today it didn't sound right.

covid_graphs/covid_graphs/scripts/country_dashboard.py
lukipuki: Yeah, but that's for version 2.1 :)

covid_graphs/covid_graphs/scripts/country_dashboard.py
lukipuki: Yeah, I thought about it, but I had to change too many things and this PR got huge anyway. Will do it some other time.

covid_graphs/covid_graphs/scripts/server.py
lukipuki: For `url_base_pathname` in Dash apps, you need to have trailing slash, otherwise it won't start. I don't know what's standard in Flask, but I assume that trailing slash is "more" correct, although less standard.

Now if you type it without the slash, there's an automatic redirect.

miskosz: That sounds good. I have not worked with Flask before btw, so also don't know what is standard.

PR #42, opened 2020-04-20
All countries on one page

lukipuki:
It's slow as hell but if that's what the users want :smile: Screenshot_20200420_082953

Will try caching the graphs next.

### DIFF
--- a/covid_graphs/covid_graphs/scripts/country_dashboard.py
+++ b/covid_graphs/covid_graphs/scripts/country_dashboard.py
@@ -1,7 +1,7 @@
 from datetime import timedelta
 from enum import Enum
 from pathlib import Path
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Tuple
 
 import dash
 import dash_core_components as dcc
@@ -23,6 +23,7 @@ class DashboardType(Enum):
 
 
 TITLE = "COVID-19 predictions of Boďová and Kollár"
+CountryGraphsByReportName = Dict[str, List[CountryGraph]]
 
 
 def _create_buttons(dashboard_type: DashboardType):
@@ -60,18 +61,19 @@ def _create_buttons(dashboard_type: DashboardType):
 
 
 def create_graphs(
-    report_by_name: Dict[str, CountryReport], prediction_event: PredictionEvent,
-):
+    report_by_short_name: Dict[str, CountryReport], prediction_event: PredictionEvent,
+) -> List[CountryGraph]:
     # Note: We silently assume there is only one prediction per country.
     country_graphs = [
-        CountryGraph(report_by_name[country_prediction.country], [country_prediction])
+        CountryGraph(report_by_short_name[country_prediction.country], [country_prediction])
         for country_prediction in prediction_db.predictions_for_event(prediction_event)
     ]
     country_graphs.sort(key=lambda graph: graph.short_name)
     return country_graphs
 
 
-def _prepare_data_structures(data_dir: Path):
+def _prepare_data_structures(data_dir: Path) -> Tuple[Dict[str, PredictionEvent],
+                                                      CountryGraphsByReportName]:
     prediction_events = prediction_db.get_prediction_events()
     prediction_events.sort(key=lambda event: event.date, reverse=True)
     prediction_event_by_name = {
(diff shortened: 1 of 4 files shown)


covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -13,11 +14,54 @@
...
+            value="Normal",
+            labelStyle={"display": "inline-block", "margin": "0 4px 0 0"},
+        ),
+    )
+    return buttons
+
+
 def create_graphs(
-    reports: List[CountryReport], prediction_event: PredictionEvent,
+    report_by_name: Dict[str, CountryReport], prediction_event: PredictionEvent,
```
miskosz: Nit: report_by_short_name

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -13,11 +14,54 @@
...
 ):
```
miskosz: Nit: You can add a return type annotation:
```
) -> List[CountryGraph]:
```

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -27,11 +71,7 @@ def create_graphs(
     return country_graphs
 
 
-def create_dashboard(
-    data_dir: Path, server: Flask,
-):
-    # TODO(miskosz): Don't use print.
-    print("Creating dashboard for prediction graphs.")
+def _prepare_data_structures(data_dir: Path):
```
miskosz: Here I would add a description of the return value, at least as a type annotation:
```
-> Tuple[Dict[str, PredictionEvent], Dict[str, List[CountryGraph]]]
```

Pro tip: You can define a type alias somewhere in the global scope at the top of the file (below imports), e.g.:
```
CountryGraphsByReportName = Dict[str, List[CountryGraph]]
```

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -53,17 +53,26 @@ def covid19_redirect():
 
 
 def _run_flask_server(server: Flask, data_dir: Path):
-    _create_prediction_app(server=server, data_dir=data_dir)
+    _create_prediction_apps(server=server, data_dir=data_dir)
```
miskosz, line 60: I admire your attention to detail 😄

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -33,7 +33,7 @@ def home():
 
     @server.route("/covid19/normal")
     def covid19_redirect():
-        return redirect(url_for("covid19_predictions"))
+        return redirect(url_for("covid19_all_predictions"))
```
miskosz, line 36: I think single predictions should be the default (i.e. the one that does not lag).

We should also redirect from `/covid19/predictions` now. (Maybe you can even have two `@server.route` decorators for one function?)

miskosz (approved):
LGTM. It got tricky to share functionality between `create_[single|all]_countr[y|ies]_dashboard`. It is good for now, we can refactor later.

covid_graphs/covid_graphs/scripts/country_dashboard.py
lukipuki: Done

covid_graphs/covid_graphs/scripts/country_dashboard.py
lukipuki: Done

covid_graphs/covid_graphs/scripts/server.py
lukipuki: I disagree, the old URL should go to the dashboard with the old functionality - "Do not break users" :smile: Yes, we might eat too much of their memory, but this consistency seems to be more important to them (the few that I talked to).

`/covid19/predictions/` can go to `/single/`, though.

covid_graphs/covid_graphs/scripts/server.py
lukipuki: My weapon and my curse :wink:

covid_graphs/covid_graphs/scripts/server.py
miskosz: Ok then either is good. Your call for the default for the `/covid19/predictions/`.

PR #43, opened 2020-04-20
Spacing between dropdowns / radios

michalburger1:
Adding some basic margin to the country/date dropdowns and scale radio buttons.

### DIFF
--- a/covid_graphs/covid_graphs/scripts/country_dashboard.py
+++ b/covid_graphs/covid_graphs/scripts/country_dashboard.py
@@ -71,9 +71,13 @@ def create_dashboard(
                 if event != OTHER
             ],
             value=BK_20200411.name,
-            style={"width": "40%"},
+            style={"width": "220px", "margin": "4px 0px"},
         ),
-        dcc.Dropdown(id="country-short-name", value="Italy", style={"width": "40%"}),
+        dcc.Dropdown(
+            id="country-short-name",
+            value="Italy",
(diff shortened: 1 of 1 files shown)


michalburger1:
Before:

before

michalburger1:
After:

after

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -71,17 +71,19 @@ def create_dashboard(
                 if event != OTHER
             ],
             value=BK_20200411.name,
-            style={"width": "40%"},
+            style={"width": "220px", "margin": "4px 0px"},
```
lukipuki, line 74: Are pixels the recommended unit even on very high definition displays? E.g. on a 4K?

And I actually prefer even higher margin, perhaps 6 pixels.

lukipuki (approved):
It seems that we found our UI expert!

covid_graphs/covid_graphs/scripts/country_dashboard.py
michalburger1: `px` doesn't actually mean pixels, it means "device independent pixels". They're defined in terms of viewing angle and they're typically implemented as 1 physical pixel on medium resolution screens and 2 physical pixels on high resolution screens.

lukipuki: TIL!

PR #45, opened 2020-04-20
Disable zooming and panning

michalburger1:
This will apparently disable all zooming and panning of the graphs. I don't think there's any reason people would want to do this and it will make the touch behavior on mobile a little more bearable (and the list of buttons much shorter).


michalburger1:
The one downside is that this changes the cursor from the cross icon to the hand icon and it looks a little weird. This behavior does not seem to be configurable, we'd have to hack our way around it.

lukipuki (approved):

PR #47, opened 2020-04-20
Remove most of the graph buttons

michalburger1:
I don't think any of them are particularly useful (what does the lasso selection do?). I'm only keeping the download one.


lukipuki (approved):
Great idea!

PR #48, opened 2020-04-20
Fix inconsistent font size on mobile

michalburger1:
The dropdowns / radio buttons render with a much smaller font size than the intro text. This is because the page will get automatically downscaled on most mobile devices and the scaling algorithm will try to compensate by adjusting the font sizes. Unfortunately, the dropdowns and radio buttons seem to be omitted from this for some reason. This change disables the text scaling altogether, meaning all the text will be equally small.


michalburger1:
63A3F0F9-56B0-48E8-A327-347876EEE929

lukipuki (changes requested):
I'd do this only with a limited width of the whole page. Something like 60 to 80 characters. Otherwise it's too wide and too small on mobile.

michalburger1:
22365868-797B-4377-B856-AB18339039FB

michalburger1:
The downside of this approach is that we *always* force the width to 750px on mobile, even on devices with larger screens (i.e. tablets). We could probably get a better solution with some more advanced CSS but not sure how to even get started with CSS in this setup.

lukipuki:
We can have a `.css` file in the `assets` directory and it would be picked up automatically by the Python framework, or so I was told. But it's much better now, this is good enough! We can still blame it on us being mostly backend engineers.

lukipuki (approved):

PR #49, opened 2020-04-20
Cache created dcc.Graph's

lukipuki:
Caching saves CPU on the server.
I've also isort-ed everything.


miskosz (commented):
Nice, LGTM.

I wonder how is it possible that the `def _prepare_data_structures` was not formatted, I'd expect CI to fail.

lukipuki:
Hahahaha, it did fail. Yesterday I pressed "Squash and Merge", expecting it to wait for the build result and then merge (it's what we had in our private GitHub at Spotify). But it didn't wait.

Screenshot_20200421_221927

PR #51, opened 2020-04-22
Test for formula.py

### LINKED ISSUE #50: covid_graphs.show_country_plot data USA doesn't work
It works with single predictions, but both at the same time break it:
Screenshot_20200420_230442

lukipuki:


covid_graphs/covid_graphs/country_report.py
```diff
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import date
```
miskosz: Minor: You can consider using just `import datetime`. I usually import from modules directly just classes, for functions I prefer calling `module.function(...)` to make it clear where does it come from. That is because classes usually start with capital letter. Here `datetime` module is actually a bit unusual, that's why I'd consider the other option. (Also it is annoying that it reserves the word `date` :) )

miskosz (commented):
The PR looks good and I would normally approve. (Also nice to add a test!) However I feel it would be better to leave this PR unmerged and do first the refactor I have in works, what do you say? (Also, I maybe should have said I am working on it.)

covid_graphs/covid_graphs/country_report.py
lukipuki: Yeah, I'm fairly new to production Python and I have been confused with the different approaches in Python, e.g. for numpy is always `import numpy as np` and then classes are imported using `from`. What you say makes sense.

lukipuki (commented):
This started as an attempt at fixing: https://github.com/lukipuki/COVID-19-simulation/issues/50

The actual fix is a one-liner in `country_graph.py`, but it ended up being a good opportunity to write a test and refactor. I'll make another PR for the changes in `country_graph.py` and will leave this open until the refactor is finished.

lukipuki:
The important fix has been moved to https://github.com/lukipuki/COVID-19-simulation/pull/52

lukipuki:
Ready for review.

I've renamed CurveConstructor to Formula. Reminds me of my original name EvaluatedFormula for Curve.

covid_graphs/covid_graphs/country_report.py
```diff
@@ -1,29 +1,52 @@
...
+    daily_active: np.ndarray
+    cumulative_active: np.ndarray
+
+
+def create_report(country_data_file: Path, short_name: str) -> CountryReport:
+    """Constructs a numpy representation of data read from 'data_dir' a given country."""
+    country_data = CountryData()
+    text_format.Parse(country_data_file.read_text(), country_data)
+
+    short_name = short_name
```
miskosz: This line seems unnecessary :)

covid_graphs/covid_graphs/country_report.py
```diff
@@ -1,29 +1,52 @@
...
+    return CountryReport(
+        short_name,
+        long_name,
+        dates,
+        daily_positive,
+        daily_dead,
+        daily_recovered,
+        daily_active,
+        cumulative_active,
+    )
```
miskosz, line 51: I prefer named calls (`short_name=short_name`) - it is in my opinion clearer and it prevents you from calls with wrong parameters when you change the function signature. In this case it feels a little silly though; it is more about style preference, so I'll leave it to you.

covid_graphs/covid_graphs/formula_test.py
```diff
@@ -0,0 +1,36 @@
...
+
+from .country_report import CountryReport
+from .formula import AtgFormula
+
+
+def test_two_curves():
+    cumulative_active = np.array([0, 0, 1, 2, 5, 18, 45])
+    start_date = datetime.date(2020, 4, 17)
+    dates = [start_date + datetime.timedelta(days=d) for d in range(len(cumulative_active))]
+    report = CountryReport("UK", "United Kingdom", dates, None, None, None, None, cumulative_active)
```
miskosz: I'd suggest to use the named params here.

covid_graphs/covid_graphs/formula_test.py
```diff
@@ -0,0 +1,36 @@
...
+    f1 = AtgFormula(tg=2, a=47, exponent=1.5, min_case_count=2)
+    # The curve starts (t=1) at 2020-04-20. The maximum of this curve is at t = TG * exponent = 3.
+    max_t1, start_date1 = 3, datetime.date(2020, 4, 20)
+
+    f2 = AtgFormula(tg=3, a=12, exponent=2, min_case_count=1)
+    # The curve starts (t=1) at 2020-04-19. The maximum of this curve is at t = TG * exponent = 6.
+    max_t2, start_date2, length2 = 6, datetime.date(2020, 4, 19), math.ceil(3 * (2 + math.sqrt(2)))
+
+    curve1 = f1.get_curve(report)
+    assert abs(curve1.y_max - (47 / 2) * (max_t1 / 2) ** 1.5 * math.exp(-max_t1 / 2)) < 1e-6
```
miskosz: `assert curve1.y_max == pytest.approx((47 / 2) * (max_t1 / 2) ** 1.5 * math.exp(-max_t1 / 2))`

miskosz (approved):
I expected that the whole universe changed, it was not that bad at all after all! :)

covid_graphs/covid_graphs/country_report.py
lukipuki: Done

covid_graphs/covid_graphs/formula_test.py
lukipuki: Done

covid_graphs/covid_graphs/country_report.py
lukipuki: Coming from the C++ world, I'm used to the Google C++ style, whose philosophy can be applied here as well.

When a function is called with an argument that doesn't have a name (i.e. a constant) or has a different name, then the original argument name is written as a comment Screenshot_20200425_072919

In Python, that would be equivalent to `CalculateProduct(values, options, completion_callback=None)`. It seems to me to be the most reasonable approach. I think that writing `CalculateProduct(values=values, options=options, completion_callback=None)` **hurts** readability, since you are adding redundant information that takes up space.

covid_graphs/covid_graphs/formula_test.py
lukipuki: Done

covid_graphs/covid_graphs/formula_test.py
```diff
@@ -0,0 +1,46 @@
...
+    f1 = AtgFormula(tg=2, a=47, exponent=1.5, min_case_count=2)
+    # The curve starts (t=1) at 2020-04-20. The maximum of this curve is at t = TG * exponent = 3.
+    max_t1, start_date1 = 3, datetime.date(2020, 4, 20)
+
+    f2 = AtgFormula(tg=3, a=12, exponent=2, min_case_count=1)
+    # The curve starts (t=1) at 2020-04-19. The maximum of this curve is at t = TG * exponent = 6.
+    max_t2, start_date2, length2 = 6, datetime.date(2020, 4, 19), math.ceil(3 * (2 + math.sqrt(2)))
+
+    curve1 = f1.get_curve(report)
+    assert pytest.approx(curve1.y_max, (47 / 2) * (max_t1 / 2) ** 1.5 * math.exp(-max_t1 / 2))
```
miskosz, line 35: This is not how `pytest.approx` works! :) See my original suggestion, we have to make a bugfix PR.
https://docs.pytest.org/en/latest/reference.html#pytest-approx

covid_graphs/covid_graphs/country_report.py
miskosz: Unfortunately readability is a function of convention, and what is worse that differs for every language... But I'm fine with the way you kept it, especially for a small project with a couple of devs like this I wouldn't be too fussy.

covid_graphs/covid_graphs/formula_test.py
lukipuki: Hahahahaha, I love how my brain just assumed that it knows the only possible API :D Fixed in 41ce34aafa939d84e9e10c2fdb82fb35f3c78f86

PR #52, opened 2020-04-22
Fix x-axis shift for multiple predictions

### LINKED ISSUE #50: covid_graphs.show_country_plot data USA doesn't work

lukipuki:
Also drop XAxisType, is not very useful.


miskosz (approved):
LGTM

PR #53, opened 2020-04-22
Refactor formula.py

miskosz:
- Merge `Formula` and `Curve` classes into one class `Curve`
- The class stores all information to be rendered (notably min and max date to render)
- The class is initialised with a function that converts a date to a float value.

I am not totally happy with how everything is implemented but I think it is a step forward. I leave for another PR the x-axis label, which should change based on GraphType.

### DIFF
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -45,9 +45,6 @@ class CountryGraph:
         self.cropped_dates = report.dates[min_start_date_idx:]
         self.cropped_cumulative_active = report.cumulative_active[min_start_date_idx:]
 
-        self.max_end_date = max(curve.end_date for curve in self.curves)
-        self.max_end_date = max(self.max_end_date, report.dates[-1] + datetime.timedelta(days=7))
-
     def create_country_figure(self, graph_type: GraphType):
 
         # Due to plotly limitations, we can only have graphs with dates on the x-axis when we
@@ -65,11 +62,10 @@ class CountryGraph:
         # Traces
         traces = []
         for color, curve in zip(colors, self.curves):
-            xs, ys = curve.get_trace(self.min_start_date, self.max_end_date)
             traces.append(
                 Scatter(
-                    x=list(map(adjust_xlabel, xs)),
-                    y=ys,
+                    x=list(map(adjust_xlabel, curve.xs)),
+                    y=curve.ys,
                     mode="lines",
                     name=curve.label,
                     line={"width": 2, "color": color},
@@ -92,14 +88,13 @@ class CountryGraph:
 
         # Add vertical dotted lines marking the maxima
         for color, curve in zip(colors, self.curves):
-            x_max, y_max = curve.get_maximum()
             shapes.append(
                 dict(
                     type="line",
-                    x0=adjust_xlabel(x_max),
+                    x0=adjust_xlabel(curve.x_max),
                     y0=1,
-                    x1=adjust_xlabel(x_max),
-                    y1=y_max,
+                    x1=adjust_xlabel(curve.x_max),
+                    y1=curve.y_max,
                     line=dict(width=2, dash="dash", color=color),
                 )
             )
@@ -146,7 +141,7 @@ class CountryGraph:
         )
         if graph_type != GraphType.Normal:
             maximal_y = max(
-                max(curve.get_maximum()[1] for curve in self.curves),
+                max(curve.y_max for curve in self.curves),
                 max(self.cropped_cumulative_active),
             )
             # Note: We silently assume `self.cropped_cumulative_active` does not contain zeros.
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -36,6 +36,7 @@ class CountryGraph:
     def __init__(
         self, report: CountryReport, country_predictions: List[CountryPrediction],
     ):
+        self.report = report
         self.short_name = country_predictions[0].country
         self.long_name = report.long_name
 
@@ -43,16 +44,14 @@ class CountryGraph:
         if len(country_predictions) >= 1:
             self.prediction_date = country_predictions[0].prediction_event.date.strftime("%Y-%m-%d")
 
-        self.graph_functions = [
-            prediction.formula.get_graph_function(country_report=report)
+        self.curves = [
+            prediction.formula.get_curve(country_report=report)
             for prediction in country_predictions
         ]
 
-        # TODO
-        self.dates = report.dates
-        self.dates_str = report.dates_str
-        self.ts = [formula.epoch_days(date) for date in self.dates]
-        self.cumulative_active = report.cumulative_active
+        # # TODO
+        # self.dates = report.dates
+        # self.cumulative_active = report.cumulative_active
 
     def create_country_figure(self, graph_type: GraphType):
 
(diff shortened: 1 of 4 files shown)


miskosz:
This PR is still WIP:

- There is an off-by-one error for graph display, for some countries the prediction is not aligned with observation (see screenshot below)
- Log-log graphs have wrong bounds

Screenshot 2020-04-22 at 23 35 47

covid_graphs/covid_graphs/formula.py
```diff
@@ -1,81 +1,77 @@
...
-        length = last_idx - start_idx + 1
-        self.y = formula.lambd(np.arange(length) + 1)
-        self.t = np.arange(length) + 1 + (start_idx - first_idx)
-        self.date_list = [
-            (first_date + timedelta(days=d)).strftime("%Y-%m-%d")
-            for d in range(start_idx, last_idx + 1)
-        ]
+        # Display values from the first day for which the number of cumulative active is
+        # more than min_case_count. This day is also "day one".
+        start_idx = np.argmax(country_report.cumulative_active >= self.min_case_count)
```
lukipuki, line 65: Katka and Rišo use a strict inequality, I think that's what's causing the shift for you. This is what we had before
```
return np.argmax(cumulative_active > formula.min_case_count)
```

Strict inequality is weird though, I'd suggest updating the predictions instead by bumping the minimum by +1.

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -118,30 +113,31 @@ def pick_xaxis_labels(object):
...
             ]
 
         figure = Figure(layout=layout)
         for color, curve in zip(colors, self.curves):
             figure.add_trace(
                 Scatter(
-                    x=pick_xaxis_labels(curve),
-                    y=curve.y,
-                    text=curve.date_list,
+                    x=list(map(adjust_xlabel, curve.xs)),
```
lukipuki: Previously, each curve was extended if it was shorter than the data of active cases. You need to take into account the length of `self.cropped_cumulative_active` when plotting it it.

covid_graphs/covid_graphs/formula.py
```diff
@@ -1,81 +1,77 @@
...
+        start_date = country_report.dates[start_idx]
+        day_zero = start_date - datetime.timedelta(days=1)
 
-        self.maximal_y = self.y.max()
-        self.maximal_idx = self.y.argmax()
+        # Display values until the second inflection point.
+        second_ip_days = math.ceil(self.tg * (self.exponent + math.sqrt(self.exponent)))
+        end_date = start_date + datetime.timedelta(days=second_ip_days)
```
lukipuki: This might be too short, the curve should end where the graph ends, so it needs to take into account the length of `cumulative_active` as well as other predictions plotted on the same graph.

miskosz:
Ready for review, I have mixed feelings though. Here is the changelog:

- `Formula` and `Curve` classes combined into a single `Curve` class
- Common date axis for all graphs, start_date is part of the formula
- Formulas work on dates, not "t"
- Fixed graph offsets (interpretation of `min_cases`)

Things that are not ideal:

- `Curve` remembers start and end date for a curve, which is then overridden by a wider range
- Weird caching of trace in `Curve`, just to be able to compute `maximal_y`.

covid_graphs/covid_graphs/formula.py
miskosz: Used `>=` according to the new specification. I understood that `min_cases` define `t=1` day. It looks like it works ok now, only `Austria` was adjusted manually in `predictions.py`.

covid_graphs/covid_graphs/formula.py
miskosz: Fixed by changing get_trace() interface.

covid_graphs/covid_graphs/predictions.py
```diff
@@ -25,146 +25,163 @@ class CountryPrediction:
...
-    CountryPrediction(
-        prediction_event=OTHER,
-        country="Slovakia",
-        formula=Formula(lambda t: 8 * t ** 1.28, r"$8 \cdot t^{1.28}$", 40, 10),
-    ),
+    # CountryPrediction(
+    #     prediction_event=OTHER,
+    #     country="Slovakia",
+    #     formula=Formula(lambda t: 8 * t ** 1.28, r"$8 \cdot t^{1.28}$", 40, 10),
+    # ),
```
miskosz, line 32: Note: I have disabled Slovakia. Do we want it? I can bring it back in another PR, but AFAIK we don't display it.

lukipuki: No, it's fine. I just had it there to draw possible exponents for my simulation code, it was really for convenience only.

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -28,77 +27,106 @@ class CountryGraph:
...
-            dict(
-                type="line",
-                x0=pick_xaxis_labels(curve)[curve.maximal_idx],
-                y0=1,
-                x1=pick_xaxis_labels(curve)[curve.maximal_idx],
-                y1=curve.maximal_y,
-                line=dict(width=2, dash="dash", color=color),
+
+        # Traces
+        traces = []
```
lukipuki, line 63: Nice restructuring!

covid_graphs/covid_graphs/formula.py
```diff
@@ -1,81 +1,90 @@
...
-    first_idx - 0-based index where the graph starts, so from cumulative_active[first_idx:].
-    last_idx - 0-based index where the graph ends
-    first_date - date corresponding to cumulative_active[0]
-    """
+    def get_curve(self, country_report: CountryReport) -> Curve:
+        label = r"$\frac{_A}{_TG} \cdot \left(\frac{t}{_TG}\right)^{_expon} / e^{t/_TG}$"
+        label = (
+            label.replace("_A", f"{self.a:.0f}")
+            .replace("_TG", f"{self.tg}")
+            .replace("_expon", f"{self.exponent}")
```
lukipuki, line 60: This is actually a hack that I quickly made. f-strings and TeX formulas don't mix well. Is there a better way than `replace`?

covid_graphs/covid_graphs/formula.py
```diff
@@ -1,81 +1,90 @@
...
+        # at least min_case_count. This day is also "day one".
+        start_idx = np.argmax(country_report.cumulative_active >= self.min_case_count)
+        start_date = country_report.dates[start_idx]
+        day_zero = start_date - datetime.timedelta(days=1)
 
-        self.maximal_y = self.y.max()
-        self.maximal_idx = self.y.argmax()
+        # Display values until the second inflection point.
+        second_ip_days = math.ceil(self.tg * (self.exponent + math.sqrt(self.exponent)))
+        end_date = start_date + datetime.timedelta(days=second_ip_days)
```
lukipuki: `end_date` is perhaps not the best name, since the curve doesn't really have an end and we might override it anyway. A possible name is `second_ip_day`, but probably not the best.

We clearly want to plot the curve at least up to the second inflection point. I'm tired and out of good ideas, but something like `display_at_least_to_date` or `min_display_end_date` would be better. I can't think of a shorter name right now.

lukipuki (approved):
Instead of caching the whole trace in `Curve`, why don't you just store `maximal_y`?

covid_graphs/covid_graphs/formula.py
miskosz: Not that I know of, maybe there is :)

miskosz:
About `maximal_y`, very good point :D I moved that code so many times I did not realise it can be simplified.

I have realised though that we actually **can** compute the plot range on initialisation, so I have simplified the Curve class even further.

PR #57, opened 2020-04-25
Join different graph types into one Dash graph

### LINKED ISSUE #56: Do not have separate Dash graphs for different graph types
Currently we have cached graphs for normal, semi-log and log-log graphs. This eats too much memory on the server. It will be a limiting factor for expanding predictions to more countries.

It also means that the full page needs to be reloaded everytime you change the graph type, so once we fix this, the UI will be more responsive.

lukipuki:
Previously we had multiple instances of Dash graphs, but this consumes
too much memory and makes the UI slower.

I've measured it and we've saved 40% memory.

There is one regression: I've dropped log-log graphs from the dashboards, since they need special treatment because of Plotly limitations. They still work in `show_country_plot` though.

### DIFF
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -155,6 +155,7 @@ class CountryGraph:
         elif graph_type == GraphType.LogLog:
             self.figure.update_xaxes(type="log", title=self.log_title)
             self.figure.update_yaxes(type="log", autorange=False, range=self.log_yrange)
+        return self.figure
 
 
 @click.command(help="COVID-19 country growth visualization")
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -38,6 +38,9 @@ class CountryGraph:
             prediction.formula.get_curve(country_report=report)
             for prediction in country_predictions
         ]
+        self.min_case_count = min(
+            prediction.formula.min_case_count for prediction in country_predictions
+        )
 
         # Crop country data to display.
         min_start_date = min(curve.start_date for curve in self.curves)
@@ -50,14 +53,11 @@ class CountryGraph:
 
         def adjust_xlabel(date: datetime.date):
             # Due to plotly limitations, we can only have graphs with dates on the x-axis when we
-            # aren't using logs.
-            # if graph_type == GraphType.Normal:
-            #     return date
-            # else:
-            #     return (date - log_xaxis_date_since).days
-
-            # Figure out a way to update this
-            return (date - log_xaxis_date_since).days
+            # x-axis isn't log-scale.
+            if graph_type != GraphType.LogLog:
+                return date
+            else:
+                return (date - log_xaxis_date_since).days
 
         colors = ["SteelBlue", "Purple", "Green"][: len(self.curves)]
 
@@ -142,9 +142,7 @@ class CountryGraph:
             math.log10(maximal_y) + 0.3,
         ]
         self.log_title = f"Days [since {log_xaxis_date_since.strftime('%b %d, %Y')}]"
-        # This doesn't reflect the reality yet
-        # self.lin_title = f"Date [starting on {log_xaxis_date_since.strftime('%b %d, %Y')}]"
-        self.lin_title = self.log_title
+        self.date_title = f"Date [starting from {self.min_case_count}th case]"
 
         self.figure = Figure(data=traces, layout=layout)
         self.update_graph_type(graph_type)
@@ -152,10 +150,10 @@ class CountryGraph:
 
     def update_graph_type(self, graph_type: GraphType):
         if graph_type == GraphType.Normal:
-            self.figure.update_xaxes(type="linear", title=self.lin_title)
+            self.figure.update_xaxes(type="date", title=self.date_title)
             self.figure.update_yaxes(type="linear", autorange=True)
         elif graph_type == GraphType.SemiLog:
-            self.figure.update_xaxes(type="linear", title=self.log_title)
+            self.figure.update_xaxes(type="date", title=self.date_title)
             self.figure.update_yaxes(type="log", autorange=False, range=self.log_yrange)
         elif graph_type == GraphType.LogLog:
             self.figure.update_xaxes(type="log", title=self.log_title)
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -13,7 +13,7 @@ from .predictions import CountryPrediction, prediction_db
 
 
 class GraphType(Enum):
-    Normal = "normal"
+    Linear = "linear"
     SemiLog = "semi-log"
     LogLog = "log-log"
 
@@ -45,7 +45,7 @@ class CountryGraph:
         self.cropped_dates = report.dates[min_start_date_idx:]
         self.cropped_cumulative_active = report.cumulative_active[min_start_date_idx:]
 
-    def create_country_figure(self, graph_type=GraphType.Normal):
+    def create_country_figure(self, graph_type=GraphType.Linear):
         log_xaxis_date_since = datetime.date(2020, 2, 1)
 
         def adjust_xlabel(date: datetime.date):
@@ -146,7 +146,7 @@ class CountryGraph:
         return self.figure
 
     def update_graph_type(self, graph_type: GraphType):
-        if graph_type == GraphType.Normal:
+        if graph_type == GraphType.Linear:
             self.figure.update_xaxes(type="date", title=self.date_title)
             self.figure.update_yaxes(type="linear", autorange=True)
         elif graph_type == GraphType.SemiLog:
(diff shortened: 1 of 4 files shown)


covid_graphs/covid_graphs/country_graph.py
```diff
@@ -45,14 +45,13 @@ def __init__(
...
-        # Due to plotly limitations, we can only have graphs with dates on the x-axis when we
-        # aren't using logs.
+    def create_country_figure(self, graph_type=GraphType.Normal):
         log_xaxis_date_since = datetime.date(2020, 2, 1)
 
         def adjust_xlabel(date: datetime.date):
-            if graph_type == GraphType.Normal:
+            # Due to plotly limitations, we can only have graphs with dates on the x-axis when we
+            # x-axis isn't log-scale.
+            if graph_type != GraphType.LogLog:
```
miskosz, line 54: If `LogLog` is disabled completely, do we need this code? Or do you plan on bringing it back?

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -117,19 +117,9 @@ def adjust_xlabel(date: datetime.date):
...
         layout = Layout(
             title=f"Active cases in {self.long_name}",
-            xaxis=dict(
-                autorange=True,
-                fixedrange=True,
-                # TODO(miskosz): Update label on radio change.
-                title=f"Date / Days [since {log_xaxis_date_since.strftime('%b %d, %Y')}]",
-                showgrid=False,
-            ),
+            xaxis=dict(autorange=True, fixedrange=True, showgrid=False,),
```
miskosz, line 120: I think that polite graphs should always have labeled axes :) What about `title="Date"`?

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -139,27 +129,32 @@ def adjust_xlabel(date: datetime.date):
...
+        maximal_y = max(
+            max(curve.y_max for curve in self.curves), max(self.cropped_cumulative_active),
+        )
+        # Note: We silently assume `self.cropped_cumulative_active` does not contain zeros.
+        self.log_yrange = [
+            math.log10(self.cropped_cumulative_active.min()) - 0.3,
+            math.log10(maximal_y) + 0.3,
+        ]
+        self.log_title = f"Days [since {log_xaxis_date_since.strftime('%b %d, %Y')}]"
+        self.date_title = f"Date [starting from {self.cropped_dates[0].strftime('%b %d, %Y')}]"
```
miskosz, line 142: I'm a bit confused now, we don't ever display "days since", do we?

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -45,14 +45,13 @@ def __init__(
         self.cropped_dates = report.dates[min_start_date_idx:]
         self.cropped_cumulative_active = report.cumulative_active[min_start_date_idx:]
 
-    def create_country_figure(self, graph_type: GraphType):
-
-        # Due to plotly limitations, we can only have graphs with dates on the x-axis when we
-        # aren't using logs.
+    def create_country_figure(self, graph_type=GraphType.Normal):
```
miskosz: Side note: `linear` might be a better name for `normal`

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -27,248 +27,269 @@ def __str__(self):
...
-        ],
-    )
-    def update_graph(prediction_event_name, graph_type_str, country_short_name):
-        graph_type = GraphType[graph_type_str]
-
-        graphs = [
-            country_graph
-            for country_graph in graph_dict[prediction_event_name]
-            if country_graph.short_name == country_short_name
+        report_by_short_name = {report.short_name: report for report in reports}
```
miskosz, line 46: You can create a dict directly:
```
report_by_short_name = {
    country_short_name: country_report.create_report(...)
    for ...
}
```

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -27,248 +27,269 @@ def __str__(self):
...
+                dcc.Graph(
+                    id="country-graph",
+                    figure=dict(layout=dict(height=700)),
+                    config=dict(modeBarButtons=[["toImage"]]),
+                )
+            ]
+        else:
+            extra_content = [html.Div(id="country-graphs")]
+
+        self.app = self._create_dash_app(dashboard_type, server, extra_content)
```
miskosz, line 65: Why did we introduce `CountryDashboard`, what state do we persist? Wouldn't it suffice for `create_country_dashboard` to return a dash app?

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -27,248 +27,269 @@ def __str__(self):
...
+            url_base_pathname=f"/covid19/predictions/{dashboard_type}/",
+            server=server,
+            external_scripts=[
+                "https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.4/MathJax.js?config=TeX-MML-AM_CHTML"
+            ],
+            meta_tags=[{"name": "viewport", "content": "width=750"}],
+        )
+        content = self._get_header_content(TITLE)
+        content += [html.Hr(), html.H1(id="graph-title")]
+        content += self._create_buttons(dashboard_type)
```
miskosz: Hm, does this actually work? I though static methods are called as `CountryDashboard. _create_buttons(...)`, but I might be wrong.

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -27,248 +27,269 @@ def __str__(self):
...
+            [Input("graph-type", "value")],
+        )
+        def update_country_graphs_20200329(graph_type_str):
+            result = []
+            for graph in self.graph_dict[BK_20200329.name]:
+                graph.update_graph_type(GraphType[graph_type_str])
+                result.append(graph.figure)
+            return result
+
+    def _get_header_content(self, title: str):
```
miskosz: It is fine to have functions outside of a class in a file in python :)

miskosz (commented):
I got a bit lost in the callback creation logic... What is the change that speeds everything up?

covid_graphs/covid_graphs/country_graph.py
lukipuki: Done

covid_graphs/covid_graphs/country_graph.py
lukipuki: It's not completely disabled, as I wrote in d277ac3, it still works outside of the server dashboard.

covid_graphs/covid_graphs/country_graph.py
lukipuki: Title is set later.

covid_graphs/covid_graphs/country_graph.py
lukipuki: We do for log-log graphs

covid_graphs/covid_graphs/scripts/country_dashboard.py
lukipuki: Yes, but `create_report` parses a proto.

covid_graphs/covid_graphs/scripts/country_dashboard.py
lukipuki: There's a bunch of helpful data structures that are kept around. I found it cumbersome to pass them around with pure functions.
The data structures are also used for the callbacks.

covid_graphs/covid_graphs/scripts/country_dashboard.py
lukipuki: Yeah, that's weird that it works :) Will change it.

covid_graphs/covid_graphs/scripts/country_dashboard.py
lukipuki: Done

lukipuki (commented):
The memory saving comes mostly from not creating so many CountryGraph's and dcc.Graph's. Previously we had 3 for each country, one for each graph type (semi-log, log-log). Now the figure is recycled and just the axes layouts are updated.

covid_graphs/covid_graphs/scripts/country_dashboard.py
miskosz: Why is that a problem?

covid_graphs/covid_graphs/scripts/country_dashboard.py
miskosz: Use should write tests, then you would find pure functions useful 😛 I'll leave it up to you.

miskosz (approved):
I did not check the logic in detail, but in general looks good to me. I would not use a class though, I think that is a C++ pattern, but again I'll leave that up to you :)

lukipuki:
I'd like to see this made in proper Python, but in the interest of time I need to move on to other more important stuff :smile:

Issue #58, opened 2020-04-25
Do not run Flask as a production server


lukipuki:
Flask's been trying to tell us something for a few weeks:
```
 * Environment: production
   WARNING: This is a development server. Do not use it in a production deployment.
   Use a production WSGI server instead.
```

I don't like running yet another app infront of nginx, so FastCGI seems to be the way to go: https://flask.palletsprojects.com/en/1.1.x/deploying/fastcgi/

rejdi:
I would say, this is the right task for Nginx in front of python backend and later for static file serving.

PR #59, opened 2020-04-25
Run server without inotify on MacOS.

miskosz:
`inotify` does not work on MacOS, but it is useful to be able to run the server for testing purposes. Detect if it is supported and run the server directly if not.


lukipuki:
We might have to run the server in a different way, see https://github.com/lukipuki/COVID-19-simulation/issues/58, so we'll see if we keep this logic around.

lukipuki (approved):

PR #60, opened 2020-04-25
Run daily fit in debug mode.

miskosz:
You can see fitted predictions by running `covid_graphs.show_country_plot` locally.
- Adjust optimisation to use 4 parametes
- We still leak a bit of display logic.

It seems to me that the optimisation is sensitive to the initialisation of `t0`. I have not investigated a lot and also have not investigated other parameters.

### DIFF
--- a/covid_graphs/covid_graphs/formula.py
+++ b/covid_graphs/covid_graphs/formula.py
@@ -6,6 +6,7 @@ from typing import Callable
 
 import numpy as np
 
+from . import fit_atg_model
 from .country_report import CountryReport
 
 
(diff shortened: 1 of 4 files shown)


covid_graphs/covid_graphs/formula.py
```diff
@@ -65,15 +66,55 @@ def get_curve(self, country_report: CountryReport) -> Curve:
...
+        return Curve(func=formula, start_date=start_date, end_date=end_date, label=label)
+
+
+def _get_display_end_date(
+    tg: float, exp: float, start_date: datetime.date, data_end_date: datetime.date
+) -> datetime.date:
+    # Display values until the second inflection point or one week after data finishes.
+    days_till_second_ip = math.ceil(tg * exp + math.sqrt(exp))
+    second_ip_date = start_date + datetime.timedelta(days=days_till_second_ip)
+    week_after_data_ends_date = data_end_date + datetime.timedelta(days=7)
```
lukipuki, line 119: I think 7 should either be a declared constant or a parameter.

covid_graphs/covid_graphs/formula.py
```diff
@@ -65,15 +66,55 @@ def get_curve(self, country_report: CountryReport) -> Curve:
...
         return Curve(func=formula, start_date=start_date, end_date=end_date, label=label)
+
+
+@dataclass
+class FittedFormula(Formula):
+    # Date until which to consider data. Inclusive.
+    until_date: datetime.date
+
+    def get_curve(self, country_report: CountryReport) -> Curve:
+        label = f"Daily fit {self.until_date.strftime('%d. %b')}"
```
lukipuki, line 89: Can you create a better label?

lukipuki (approved):

PR #62, opened 2020-04-26
Do not store the trace inside the curve

lukipuki:
Curve is a very simple dataclass that can generate its trace on demand.

I've also added PolynomialFormula, since it was useful for data
exploration and blog posts.

I've fixed a few bugs, including wrong inflection point calculation.

Screenshot_20200426_074159

### DIFF
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -12,6 +12,8 @@ from .country_report import CountryReport, create_report
 from .formula import FittedFormula
 from .predictions import CountryPrediction, PredictionEvent, prediction_db
 
+EXTENSION_PERIOD = datetime.timedelta(days=7)
+
 
 class GraphType(Enum):
     Linear = "linear"
@@ -45,9 +47,13 @@ class CountryGraph:
         min_start_date_idx = report.dates.index(min_start_date)
         self.cropped_dates = report.dates[min_start_date_idx:]
         self.cropped_cumulative_active = report.cumulative_active[min_start_date_idx:]
+        self.display_at_least_until = max(
+            max(curve.display_at_least_until for curve in self.curves) + datetime.timedelta(days=1),
+            self.cropped_dates[-1] + EXTENSION_PERIOD,
+        )
 
     def create_country_figure(self, graph_type=GraphType.Linear):
-        log_xaxis_date_since = datetime.date(2020, 2, 1)
+        log_xaxis_date_since = self.cropped_dates[0] - datetime.timedelta(days=1)
 
         def adjust_xlabel(date: datetime.date):
             # Due to plotly limitations, we can only have graphs with dates on the x-axis when we
@@ -59,17 +65,32 @@ class CountryGraph:
 
         colors = ["SteelBlue", "Purple", "Green", "Orange", "Gray"][: len(self.curves)]
 
-        # Traces
-        traces = []
+        traces, shapes = [], []
+        maximal_y = max(self.cropped_cumulative_active)
         for color, curve in zip(colors, self.curves):
+            xs, ys = curve.generate_trace(self.display_at_least_until)
+            idx_max = ys.argmax()
+            maximal_y = max(maximal_y, ys.max())
+
             traces.append(
                 Scatter(
-                    x=list(map(adjust_xlabel, curve.xs)),
-                    y=curve.ys,
-                    text=curve.xs,
+                    x=list(map(adjust_xlabel, xs)),
+                    y=ys,
+                    text=xs,
                     mode="lines",
                     name=curve.label,
                     line={"width": 2, "color": color},
+                )
+            )
+            # Add vertical dotted line marking the maximum
+            shapes.append(
+                dict(
+                    type="line",
+                    x0=adjust_xlabel(xs[idx_max]),
+                    y0=1,
+                    x1=adjust_xlabel(xs[idx_max]),
+                    y1=ys.max(),
+                    line=dict(width=2, dash="dash", color=color),
                 )
             )
 
@@ -83,22 +104,6 @@ class CountryGraph:
                 line=dict(width=3, color="rgb(239, 85, 59)"),
             )
         )
-
-        # Shapes
-        shapes = []
-
-        # Add vertical dotted lines marking the maxima
-        for color, curve in zip(colors, self.curves):
-            shapes.append(
-                dict(
-                    type="line",
-                    x0=adjust_xlabel(curve.x_max),
-                    y0=1,
-                    x1=adjust_xlabel(curve.x_max),
-                    y1=curve.y_max,
-                    line=dict(width=2, dash="dash", color=color),
-                )
-            )
 
         # Add green zone marking the data available at the prediction date.
         shapes.append(
@@ -131,12 +136,9 @@ class CountryGraph:
             plot_bgcolor="White",
         )
 
-        maximal_y = max(
-            max(curve.y_max for curve in self.curves), max(self.cropped_cumulative_active),
-        )
         # Note: We silently assume `self.cropped_cumulative_active` does not contain zeros.
         self.log_yrange = [
-            math.log10(self.cropped_cumulative_active.min()) - 0.3,
+            math.log10(max(0.1, self.cropped_cumulative_active.min())) - 0.3,
             math.log10(maximal_y) + 0.3,
         ]
         self.log_title = f"Days [since {log_xaxis_date_since.strftime('%b %d, %Y')}]"
@@ -171,7 +173,6 @@ def show_country_plot(data_dir: Path, country_name: str):
     country_report = create_report(data_dir / f"{country_name}.data", short_name=country_name)
 
     for until_date in [
-        datetime.date(2020, 3, 29),
         datetime.date(2020, 4, 11),
         country_report.dates[-1],
     ]:
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -2,14 +2,14 @@ import datetime
 import math
 from enum import Enum
 from pathlib import Path
-from typing import List
+from typing import List, Tuple
 
 import click
 import click_pathlib
 from plotly.graph_objs import Figure, Layout, Scatter
 
 from .country_report import CountryReport, create_report
-from .formula import FittedFormula
+from .formula import Curve, FittedFormula
 from .predictions import CountryPrediction, PredictionEvent, prediction_db
 
 EXTENSION_PERIOD = datetime.timedelta(days=7)
@@ -22,6 +22,17 @@ class GraphType(Enum):
 
     def __str__(self):
         return self.value
+
+
+def _get_display_range(
+    report: CountryReport, curves: List[Curve]
+) -> Tuple[datetime.date, datetime.date]:
+    start_date = min(curve.start_date for curve in curves)
+
+    display_until = max(
+        max(curve.display_at_least_until for curve in curves), report.dates[-1] + EXTENSION_PERIOD,
+    )
+    return start_date, display_until
 
 
 class CountryGraph:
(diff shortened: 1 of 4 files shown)


covid_graphs/covid_graphs/country_graph.py
```diff
@@ -12,6 +12,8 @@
 from .formula import FittedFormula
 from .predictions import CountryPrediction, PredictionEvent, prediction_db
 
+EXTENSION_PERIOD = datetime.timedelta(days=7)
```
miskosz, line 15: I think we should extend this to 14 days. Otherwise the prediction ends too soon after data (if after 2nd IP) and does not tell much.

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -45,9 +47,13 @@ def __init__(
         min_start_date_idx = report.dates.index(min_start_date)
         self.cropped_dates = report.dates[min_start_date_idx:]
         self.cropped_cumulative_active = report.cumulative_active[min_start_date_idx:]
+        self.display_at_least_until = max(
+            max(curve.display_at_least_until for curve in self.curves) + datetime.timedelta(days=1),
+            self.cropped_dates[-1] + EXTENSION_PERIOD,
+        )
```
miskosz: Why do you add 1 day to `curve.display_at_least_until `?

Also here I'd call it rather `self.display_until` since it is definitive. I think that for consistency, we should calculate also `display_from` here.

miskosz: On another note, if the fit is crazy I think it makes sense to limit `self.display_until ` to be at most last data date plus 2 months (can do also another PR).

I'd also suggest to extract the functionality to a separate function `_get_display_range(curves, report.dates)`, it has become non-trivial.

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -45,9 +47,13 @@ def __init__(
...
 
     def create_country_figure(self, graph_type=GraphType.Linear):
-        log_xaxis_date_since = datetime.date(2020, 2, 1)
+        log_xaxis_date_since = self.cropped_dates[0] - datetime.timedelta(days=1)
```
miskosz, line 63: Did you test this? I think for some countries with small values (like Jordan I think) offsetting the x values by more looked better.

miskosz: Sidenote: In the future I can imagine more countries in one graph; having consistent axis will be then necessary. (But not needed for now.)

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -59,19 +65,34 @@ def adjust_xlabel(date: datetime.date):
...
         colors = ["SteelBlue", "Purple", "Green", "Orange", "Gray"][: len(self.curves)]
 
-        # Traces
-        traces = []
+        traces, shapes = [], []
+        maximal_y = max(self.cropped_cumulative_active)
         for color, curve in zip(colors, self.curves):
+            xs, ys = curve.generate_trace(self.display_at_least_until)
+            idx_max = ys.argmax()
+            maximal_y = max(maximal_y, ys.max())
```
miskosz, line 80: I think this is a bug, `maximal_y` gets redefined as we iterate through curves. I suggest to rename on L69 to `max_cumulative_active`.

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -59,19 +65,34 @@ def adjust_xlabel(date: datetime.date):
 
         colors = ["SteelBlue", "Purple", "Green", "Orange", "Gray"][: len(self.curves)]
 
-        # Traces
-        traces = []
+        traces, shapes = [], []
```
miskosz, line 75: I was thinking we should maybe make `_get_traces()` and `_get_shapes()` functions. If we decide to do that we can do another PR.

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -131,12 +136,9 @@ def adjust_xlabel(date: datetime.date):
             plot_bgcolor="White",
         )
 
-        maximal_y = max(
-            max(curve.y_max for curve in self.curves), max(self.cropped_cumulative_active),
-        )
         # Note: We silently assume `self.cropped_cumulative_active` does not contain zeros.
         self.log_yrange = [
-            math.log10(self.cropped_cumulative_active.min()) - 0.3,
+            math.log10(max(0.1, self.cropped_cumulative_active.min())) - 0.3,
```
miskosz: I strongly suggest to use `1.0` instead of `0.1`. Otherwise the lower bound can be negative, and the graph will stretch infinitely long near zero.... I'm surprised if the displayed graph does not break actually.

covid_graphs/covid_graphs/fit_atg_model.py
```diff
@@ -13,7 +13,7 @@ class AtgModelFit:
     t0: float
 
     def predict(self, x: float) -> float:
-        ys = _model(params=[self.a, self.tg, self.exp, self.t0], xs=np.array([x]))
+        ys = _model(params=[self.a, self.tg, self.exp, 0], xs=np.array([x]))
```
miskosz, line 16: Lol, how is this possible :D I should have tested the function.

covid_graphs/covid_graphs/formula.py
```diff
@@ -2,42 +2,27 @@
...
-            start_date + datetime.timedelta(days=d) for d in range((end_date - start_date).days)
-        ]
-        self.ys = [func(x) for x in self.xs]
-
-        idx_max = np.argmax(self.ys)
-        self.x_max = self.xs[idx_max]
-        self.y_max = self.ys[idx_max]
+    func: Callable
+    start_date: datetime.date
+    display_at_least_until: datetime.date
```
miskosz, line 17: I find the dissymmetry between start an end suspicious. Is there a reason? What about using `display_at_least_from`? The function `generate_trace` would then take `start_date` as well.

covid_graphs/covid_graphs/formula.py
```diff
@@ -2,42 +2,27 @@
...
+    label: str
+
+    def generate_trace(self, end_date: datetime.date) -> Tuple[List[datetime.date], np.ndarray]:
+        """Generates trace corresponding to the half-open interval [self.start_date, end_date)"""
+        raw_xs = np.arange((end_date - self.start_date).days)
```
miskosz: Oh, you changed `func` to be `float -> float`. I think this is a step back in the design, in my opinion the logic of how to convert `date` to `x` values should be the responsibility of Formula generator, and should not leak here. Is there a reason for this change?

covid_graphs/covid_graphs/formula.py
```diff
@@ -46,6 +31,41 @@ def get_curve(self, country_report: CountryReport) -> Curve:
...
+
+        return Curve(
+            func=lambda x: self.a * (x ** self.exponent),
+            start_date=start_date,
+            display_at_least_until=display_at_least_until,
+            label=label,
+        )
+
+
+def create_label(a: float, tg: float, exponent: float, prefix=""):
```
miskosz: Maybe `_create_atg_label`? (In python, the convention is to prepend `_` for functions that are not supposed to be used outside of the current file.)

covid_graphs/covid_graphs/formula.py
miskosz: I thought about it a bit more. The real question here is who should be aware of how to convert date to float.

miskosz (changes requested):
Nice PR! I noticed though some things that we should resolve before merging :)

covid_graphs/covid_graphs/country_graph.py
lukipuki: I did test it. I needed a nice graph for the post for `r/math` and needed a cubic polynomial to be a straight line.
italy-log-log

Since we aren't showing multiple countries and there's a real need for this functionality, I'd keep it as is for now.

covid_graphs/covid_graphs/country_graph.py
lukipuki: It's kind of strange to argue about arbitrary constants. Also, I'd prefer if this was configurable and not a constant. I'll add a TODO, since it wasn't me who introduced 7 and it's not what this PR is about.

covid_graphs/covid_graphs/country_graph.py
lukipuki: We're calculating a non-inclusive date here, so we're adding 1. However, that's probably confusing and difficult to name properly, so I removed the +1 and changed the API of `create_trace` instead.

Also,
- I've renamed `self.display_at_least_until` to `self.display_until`
- I've introduced `_get_display_range`

covid_graphs/covid_graphs/country_graph.py
lukipuki: Sure, another PR :)

covid_graphs/covid_graphs/country_graph.py
lukipuki: Redefined? It's just a variable update.

covid_graphs/covid_graphs/country_graph.py
lukipuki: This was just a quick fix to avoid a run-time error, I didn't check the display. Set to 1.0 now.

covid_graphs/covid_graphs/fit_atg_model.py
lukipuki: This is not a bug, it's a consequence of me changing (simplifying) the API.

Actually, I think the design of `fit_atg_model.py` could be modified more to fit the simpler API, but I haven't spent too much time thinking about it.

covid_graphs/covid_graphs/formula.py

covid_graphs/covid_graphs/fit_atg_model.py
miskosz: Oh, now I get it. This seems to me like some logic is leaking...

covid_graphs/covid_graphs/formula.py
```diff
@@ -95,26 +110,25 @@ def get_curve(self, country_report: CountryReport) -> Curve:
...
+            tg=fit.tg, exp=fit.exp, start_date=start_date,
         )
 
-        def formula(date: datetime.date):
-            x = (date - date_zero).days
-            return fit.predict(x)
-
-        return Curve(func=formula, start_date=start_date, end_date=end_date, label=label)
+        return Curve(
+            func=lambda x: fit.predict(x),
```
miskosz, line 129: Re: Logic leaking in `fit_atg_model`: Now we are passing different `x` in then we were before. We should keep it consistent. Also we should be careful, `fit.t0` can be non-integral.

covid_graphs/covid_graphs/formula.py
lukipuki: What's the point of showing polynomial values such as (-4)^6.23? The polynomial has a start at t=0. That's what explains the asymmetry. I've added a class comment:

>     Curve is created from a function `func(x)`, which has x=0 at `start_date`. It's only defined for
>     x >= 0.

covid_graphs/covid_graphs/formula.py
lukipuki: The previous API was convoluted and prone to errors. Even the author himself made some off-by-one errors that I've fixed :smile: 

We had to calculate `(date - day_zero).days` at three different places for no good reason. Now it's simpler:

- We define a function, e.g. `x^2 / e^(x/5)`
- We say where `x=0` is (i.e. set `start_date`).
- Profit! Fewer bugs and everyone is happier.

There's no reason to convert dates to function values at many places, the only place this is done is in `generate_trace`. Much cleaner IMO!

covid_graphs/covid_graphs/country_graph.py
miskosz: Ah, sure, nevermind :)

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -173,7 +180,8 @@ def show_country_plot(data_dir: Path, country_name: str):
     country_report = create_report(data_dir / f"{country_name}.data", short_name=country_name)
 
     for until_date in [
-        datetime.date(2020, 4, 11),
+        country_report.dates[-12],
+        country_report.dates[-7],
```
miskosz, line 184: Smart :)

covid_graphs/covid_graphs/formula.py
```diff
@@ -17,9 +17,14 @@ class Curve:
...
 
+    """
+    Curve is created from a function `func(x)`, which has x=0 at `start_date`. It's only defined for
+    x >= 0.
+    """
+
     def generate_trace(self, end_date: datetime.date) -> Tuple[List[datetime.date], np.ndarray]:
-        """Generates trace corresponding to the half-open interval [self.start_date, end_date)"""
-        raw_xs = np.arange((end_date - self.start_date).days)
+        """Generates trace corresponding to the closed interval [self.start_date, end_date]"""
```
miskosz, line 26: This is a good change.

miskosz (approved):
Looks good. I think it is much simpler with the narrowed down the meaning of `Curve`, when we clarified that :) I can make the fix for `t0` in another PR.

covid_graphs/covid_graphs/country_graph.py
lukipuki: :grinning: Yeah, I was trying something. Do you want me to revert this? I do find it useful.

PR #63, opened 2020-04-26
Fix issue with float t0.

miskosz:


covid_graphs/covid_graphs/country_graph.py
```diff
@@ -168,6 +168,23 @@ def update_graph_type(self, graph_type: GraphType):
...
+            ),
+            country=report.short_name,
+            formula=FittedFormula(until_date=until_date),
+        )
+        for until_date in [
+            report.dates[-12],
+            report.dates[-7],
+            report.dates[-1],
+        ]
+    ]
```
miskosz, line 181: I've pulled this out for easier testing.

covid_graphs/covid_graphs/fit_atg_model.py
```diff
@@ -26,7 +26,7 @@ def fit_atg_model(xs: np.ndarray, ys: np.ndarray) -> AtgModelFit:
     a_init = 2000.0
     tg_init = 7.0
     exp_init = 6.23
-    t0_init = 0.0
+    t0_init = xs[0]
```
miskosz, line 29: This makes the optimisation independent of the chosen conversion of dates to floats.

lukipuki (approved):
Nice! The fractional shift is visible, when I compare with a graph from `master`.

PR #65, opened 2020-04-27
Update README regarding Conan

lukipuki:
* Use Conan virtualrunenv
* Conan automatic profile detection

(Most of you don't care about the C++ simulation, but you can at least check if the instructions make sense).

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,33 +20,42 @@ The simulation is for the initial phase of the epidemic in Slovakia, based on [S
 
 ## Building using Conan
 
-Conan is able to download the dependencies and compile the project. However, you still need OpenMP on your system, though that usually comes installed with the compiler (for Clang, you need to also install `libomp`).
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -20,33 +20,42 @@ The simulation is for the initial phase of the epidemic in Slovakia, based on [S
...
+# Alternatively, adding '-o growth_type=exponential' creates a binary with simulated exponential growth.
+```
 
 ## Running and viewing protobuf files
 
 ```sh
-COVID-19-simulation/bin/simulation data/Slovakia.data
-# The simulation creates a proto file 'polynomial.sim' or 'exponential.sim', which can be examined
+conan install -g virtualrunenv 'COVID-19-simulation/1.0.0@'
+source activate_run.sh
```
miskosz, line 42: Where does `activate_run.sh` come from? Is it created by Conan?

miskosz (approved):
LGTM

README.md
lukipuki: Yes https://docs.conan.io/en/latest/reference/generators/virtualrunenv.html#usage

PR #67, opened 2020-04-27
REST service for data & predictions

rejdi:


covid_graphs/covid_graphs/scripts/rest.py
```diff
@@ -0,0 +1,105 @@
...
+            graph = CountryGraph(report=country_report, country_predictions=country_predictions)
+            country_reports_active[country] = {
+                "type": "cumulative_active",
+                "date_list": graph.cropped_dates,
+                "values": graph.cropped_cumulative_active.tolist(),
+                "short_name": country_report.short_name,
+                "long_name": country_report.long_name,
+            }
+            for prediction, curve in zip(country_predictions, graph.curves):
+                xs, ys = curve.generate_trace(graph.display_until)
```
lukipuki, line 62: Our API is insufficient, you shouldn't be calling `generate_trace` here. We should fix it in https://github.com/lukipuki/COVID-19-simulation/issues/68

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -71,9 +72,26 @@ def covid19_predictions_redirect():
...
 
 
+def _create_rest(data_dir: Path, server: Flask):
+    rest = Rest(data_dir=data_dir)
+
+    @server.route("/covid19/data/<country>")
+    def covid19_country_data(country):
+        return rest.get_country_data(country)
+
+    @server.route("/covid19/predictions/list/")
```
lukipuki, line 86: Is it common to mix user visible API with internal API?

`/covid19/predictions/all` is user-facing, while `/covid19/predictions/list` is for internal use. I find it odd. Could we add `/rest` prefix somewhere?

Also, I'm worried someone will discover this API and fetch JSON data from us. Is it possible to only bind it to the localhost? Perhaps some nginx magic is needed here.

lukipuki (commented):
By REST do you mean the actual REST https://en.wikipedia.org/wiki/Representational_state_transfer or is it a synonym of HTTP+JSON?

covid_graphs/covid_graphs/scripts/server.py
rejdi: 1. `/rest`, sure.
1. hide JSON data - no. It's by design, because UI logic runs completely in browser. BTW: current implementation with dash is already sending graph data using HTTP+JSON. It's not that nice as this api, but it's still possible to fetch that data.

covid_graphs/covid_graphs/scripts/rest.py
rejdi: So, does this PR have to wait on #68?

rejdi:
According to wikipedia definition, I'd say it's more HTTP+JSON than REST, almost because our resources are read-only. However most attributes applies to this service.

covid_graphs/covid_graphs/scripts/server.py
lukipuki: Right

covid_graphs/covid_graphs/scripts/rest.py
lukipuki: No, we'll fix this part once #68 is done

lukipuki (approved):
Wait, read-only API isn't REST? Which part is not fulfilled?

rejdi:
Except read-only part I think - Hypermedia as the engine of application state. In our case it would mean adding links into `/covid19/predictions/list/` for each prediction. Moreover there should be endpoint for listing all available country data. For now, only we know, that under `/covid19/data/<country>` is something and there is no explicit list of what `<country>` might be.

covid_graphs/covid_graphs/scripts/rest.py
```diff
@@ -0,0 +1,105 @@
...
+from covid_graphs.predictions import prediction_db
+
+TITLE = "COVID-19 predictions of Boďová and Kollár"
+CountryGraphsByReportName = Dict[str, List[CountryGraph]]
+
+
+class Rest:
+    def __init__(self, data_dir: Path):
+        # vytvor strukturu pre predikcie a krajiny
+        # vytvor strukturu pre data
```
miskosz, line 17: Nit: We try to keep the codebase in English :)

covid_graphs/covid_graphs/scripts/rest.py
```diff
@@ -0,0 +1,105 @@
...
+            if (data_dir / f"{country_short_name}.data").is_file():
+                country_report = create_report(
+                    data_dir / f"{country_short_name}.data", country_short_name
+                )
+                country_reports[country_report.short_name] = country_report
+
+        return country_reports
+
+    @staticmethod
+    def _create_predictions(country_reports):
```
miskosz, line 47: Although we don't enforce types, the typecheck is more useful if they are specified:
```
def _create_predictions(country_reports: List[CountryReport]) -> Tuple[List[Dict], Dict[str, Dict]]:
```
Same for the missing return types in the functions above.

covid_graphs/covid_graphs/scripts/rest.py
```diff
@@ -0,0 +1,105 @@
...
+                        "max_value": ys.max(),
+                        "date_name": prediction.prediction_event.name,
+                        "date": prediction.prediction_event.date,
+                    }
+                )
+
+        return predictions, country_reports_active
+
+    def get_app(self):
+        return self
```
miskosz, line 82: What is this useful for?

miskosz (commented):
Left a comment or two :)

covid_graphs/covid_graphs/scripts/rest.py
rejdi: Keeping same convention as is in CountryDashboard

miskosz: I don't think a lot of thought went into naming of CountryDashboard methods :) (Honestly, there is already a confusion between flask app and dash app so... yeah.)

PR #69, opened 2020-04-27
Change rest endpoints.

rejdi:
Changed rest endpoints on @lukipuki request.


lukipuki (approved):

miskosz:
Don't you want to address my comments as well since we are at it, or do you plan another PR? (Or don't plan at all? :) )

rejdi:
Some do, some don't ;). Will try now.

PR #70, opened 2020-04-27
Separate displaying logic from curve/trace computations

### LINKED ISSUE #68: Refactor country_graph.py for better integration into other graphs
The part inside `__init__` should be stored in a dataclass: https://github.com/lukipuki/COVID-19-simulation/blob/b3c8ec6a35a300632fdb6fed96313d9dbec58fa3/covid_graphs/covid_graphs/country_graph.py#L38-L60

Curves should be stored inside the dataclass too, together with their traces (or rather just their traces).

The function `create_country_figure` shouldn't contain any computation, just display configuration.

lukipuki:
I'm now using the synonym trace, since it's a nice sign that we switched the API. There isn't much else behind the name change from `Curve` to `Trace`.

### DIFF
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -1,13 +1,11 @@
 import datetime
 import math
-from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
 from typing import List, Tuple
 
 import click
 import click_pathlib
-import numpy as np
 from plotly.graph_objs import Figure, Layout, Scatter
 
 from .country_report import CountryReport, create_report
@@ -26,22 +24,6 @@ class GraphType(Enum):
         return self.value
 
 
-@dataclass
-class Trace:
-    xs: List[datetime.date]
-    ys: np.ndarray
-    x_max: datetime.date
-    y_max: float
-    label: str
-
-
-def _create_trace(trace_generator: TraceGenerator, display_until: datetime.date) -> Trace:
-    xs, ys = trace_generator.generate_trace(display_until)
-    idx_max = ys.argmax()
-    x_max, y_max = xs[idx_max], ys[idx_max]
-    return Trace(xs, ys, x_max, y_max, label=trace_generator.label)
-
-
 def _get_display_range(
     report: CountryReport, trace_generators: List[TraceGenerator]
 ) -> Tuple[datetime.date, datetime.date]:
@@ -54,7 +36,6 @@ def _get_display_range(
     return start_date, display_until
 
 
-# @dataclass
 class CountryGraph:
     """Constructs a graph for a given country"""
 
@@ -83,7 +64,7 @@ class CountryGraph:
         start_date, display_until = _get_display_range(report, trace_generators)
 
         self.traces = [
-            _create_trace(trace_generator, display_until) for trace_generator in trace_generators
+            trace_generator.generate_trace(display_until) for trace_generator in trace_generators
         ]
 
         start_date_idx = report.dates.index(start_date)
@@ -92,14 +73,14 @@ class CountryGraph:
         self.cropped_cumulative_active = report.cumulative_active[start_date_idx:]
         self.log_xaxis_date_since = self.cropped_dates[0] - datetime.timedelta(days=1)
         self.log_title = f"Days [since {self.log_xaxis_date_since.strftime('%b %d, %Y')}]"
-        self.date_title = f"Date [starting from {self.cropped_dates[0].strftime('%b %d, %Y')}]"
+        self.date_title = "Date"
 
-        maximal_y = max(
-            max(self.cropped_cumulative_active), max(trace.y_max for trace in self.traces)
+        max_value = max(
+            max(self.cropped_cumulative_active), max(trace.max_value for trace in self.traces)
         )
         self.log_yrange = [
             math.log10(max(1, self.cropped_cumulative_active.min())) - 0.3,
-            math.log10(maximal_y) + 0.3,
+            math.log10(max_value) + 0.3,
         ]
 
     def create_country_figure(self, graph_type=GraphType.Linear):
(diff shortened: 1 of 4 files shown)


covid_graphs/covid_graphs/country_graph.py
```diff
@@ -24,17 +26,35 @@ def __str__(self):
         return self.value
 
 
+@dataclass
+class Trace:
+    xs: List[datetime.date]
+    ys: np.ndarray
+    x_max: datetime.date
+    y_max: float
```
miskosz: Note: I like @rejdi's naming for the rest server: `max_value` and `max_value_date`. Not completely convinced if suitable here but you can decide.

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -24,17 +26,35 @@ def __str__(self):
...
 
     display_until = max(
-        max(curve.display_at_least_until for curve in curves), report.dates[-1] + EXTENSION_PERIOD,
+        max(trace_generator.display_at_least_until for trace_generator in trace_generators),
+        report.dates[-1] + EXTENSION_PERIOD,
     )
     return start_date, display_until
 
 
+# @dataclass
```
miskosz: Forgotten line?

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -44,59 +64,75 @@ def __init__(
         self.short_name = report.short_name
         self.long_name = report.long_name
 
-        # TODO(lukas): Figure out better strategy for more predictions
         if len(country_predictions) >= 1:
-            self.prediction_date = country_predictions[0].prediction_event.date
+            self.prediction_date = max(
+                country_prediction.prediction_event.date
+                for country_prediction in country_predictions
+            )
```
miskosz, line 52: Note: This can stay as is. I think though that `self.prediction_date` is an ill-defined concept. If we have more than one prediction date we have to come up with a new UI to be able to display them per trace.

Actually, I'd move this logic just before plotting the green zone, there is no need to compute it at init.

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -44,59 +64,75 @@ def __init__(
...
-        min_start_date_idx = report.dates.index(min_start_date)
+        start_date_idx = report.dates.index(start_date)
         # Crop country data to display.
-        self.cropped_dates = report.dates[min_start_date_idx:]
-        self.cropped_cumulative_active = report.cumulative_active[min_start_date_idx:]
+        self.cropped_dates = report.dates[start_date_idx:]
+        self.cropped_cumulative_active = report.cumulative_active[start_date_idx:]
+        self.log_xaxis_date_since = self.cropped_dates[0] - datetime.timedelta(days=1)
+        self.log_title = f"Days [since {self.log_xaxis_date_since.strftime('%b %d, %Y')}]"
+        self.date_title = f"Date [starting from {self.cropped_dates[0].strftime('%b %d, %Y')}]"
```
miskosz: I'd suggest `self.date_title = "Date"`, "starting from" does not make really sense.

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -24,17 +26,35 @@ def __str__(self):
...
+@dataclass
+class Trace:
+    xs: List[datetime.date]
+    ys: np.ndarray
+    x_max: datetime.date
+    y_max: float
+    label: str
+
+
+def _create_trace(trace_generator: TraceGenerator, display_until: datetime.date) -> Trace:
```
miskosz: This function and `TraceGenerator.generate_trace` can be merged:
* `Trace` moves to formula.py
* `TraceGenerator.generate_trace` returns a `Trace`

That seems to me more logical.

covid_graphs/covid_graphs/formula_test.py
```diff
@@ -4,11 +4,12 @@
 import numpy as np
 import pytest
 
+from .country_graph import _create_trace
```
miskosz: It is weird to have this test to depend on `country_graph`, but this should be solved if the trace creation functionality is moved.

miskosz (changes requested):
Mostly LGTM, I have a suggestion about the trace generating API.

miskosz:
Also I like the name `Trace` better than `Curve` :)

covid_graphs/covid_graphs/country_graph.py
lukipuki: Done

covid_graphs/covid_graphs/country_graph.py
lukipuki: Done

covid_graphs/covid_graphs/country_graph.py
lukipuki: Done

covid_graphs/covid_graphs/country_graph.py
lukipuki: Done (for context, we used to have number of cases in the sentence)

covid_graphs/covid_graphs/formula_test.py
lukipuki: Done

covid_graphs/covid_graphs/country_graph.py
lukipuki: Yes, we don't have a good UI for this, but taking a maximum is a much better choice.

The point of this refactoring is to do all computation before the UI is drawn, so that it can be also reused in the JS web. We should store it here, otherwise we have to keep all predictions and compute it from them later.

miskosz: I would argue that UI display should get all the prediction dates, otherwise we are losing information. (It is a minor comment, I'm fine with not doing that.)

miskosz (approved):
Nice refactor, lgtm!

covid_graphs/covid_graphs/country_graph.py
lukipuki: Sure, but that's a refactor that I don't want to do right now :) I just wanted to refactor and keep the same functionality.

PR #71, opened 2020-04-27
Create a page for fitted country predictions.

miskosz:
The file `country_dashboard.py` needs a refactor, I did not do it because we will hopefully deprecate the file soon with The New Web™.

I think we should merge this only if:

- Memory is not an issue
- We want to provide users value (show predictions for recent dates) before we implement a proper solution (generate predictions with data import)

### DIFF
--- a/covid_graphs/covid_graphs/scripts/country_dashboard.py
+++ b/covid_graphs/covid_graphs/scripts/country_dashboard.py
@@ -10,6 +10,7 @@ from dash.dependencies import Input, Output
 from flask import Flask
 
 import covid_graphs.country_report as country_report
+from covid_graphs import country_graph
 from covid_graphs.country_graph import CountryGraph, GraphType
 from covid_graphs.country_report import CountryReport
 from covid_graphs.predictions import BK_20200329, BK_20200411, OTHER, PredictionEvent, prediction_db
@@ -17,6 +18,7 @@ from covid_graphs.predictions import BK_20200329, BK_20200411, OTHER, Prediction
 
 class DashboardType(Enum):
     SingleCountry = "single"
+    SingleCountryAllPredictions = "country"
     AllCountries = "all"
 
     def __str__(self):
@@ -35,7 +37,6 @@ class CountryDashboard:
             prediction_event.name: prediction_event for prediction_event in prediction_events
         }
 
-        # TODO: parsing the proto can fail
         reports = [
             country_report.create_report(
                 data_dir / f"{country_short_name}.data", country_short_name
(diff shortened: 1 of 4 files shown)


covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -104,6 +107,10 @@ def _create_prediction_apps(data_dir: Path, server: Flask):
     def covid19_single_predictions():
         return single_prediction_app.index()
 
+    @server.route("/covid19/predictions/country/")
```
lukipuki: This is a strange endpoint name. We have `/all` and `/single`, and `/country` to me sounds like synonymous with `/single`. 
```suggestion
    @server.route("/covid19/predictions/automatic/")
```

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -185,6 +216,26 @@ def update_graph(prediction_event_name, graph_type_str, country_short_name):
             figure = graphs[0].create_country_figure(graph_type)
             return figure
 
+    def _create_single_country_all_predictions_callbacks(self):
+        # TODO: Predictions will be generated statically in the future.
```
lukipuki, line 220: It's fine doing it at the restart (it might become a problem with 100 countries though).

covid_graphs/covid_graphs/scripts/country_dashboard.py
```diff
@@ -84,19 +92,30 @@ def _create_graphs(
...
+        if (
+            dashboard_type == DashboardType.SingleCountry
+            or dashboard_type == DashboardType.AllCountries
+        ):
+            buttons.append(
+                dcc.Dropdown(
+                    id="prediction-event",
+                    options=[
+                        dict(
+                            label=(event.date + timedelta(days=1)).strftime("%B %d"),
```
lukipuki, line 108: I feel like we've been talking about adding this field since 2018, way before COVID-19 :grinning: 
```suggestion
                            label=(event.creation_date).strftime("%B %d"),
```

lukipuki (approved):
I'm really motivated to publish some automated predictions soon, ideally before the weekend or on Saturday, so let's merge this even if it means we need to buy some server memory (it looks like we don't, though).

Reworking the web to the new version with a constantly changing backend and us working full-time jobs will take time (paging @rejdi ), so we'll have to live with `country_dashboard.py` for a while. I have actually started refactoring it in hopes of reducing memory footprint further - it looks like it didn't help but the code is nicer :smile: 

Before linking this from `index.html`, we should improve the labels (and probably also drop the very first prediction). Could be another PR.
Screenshot_20200428_001944

covid_graphs/covid_graphs/scripts/server.py
miskosz: What do you think about `daily`?

covid_graphs/covid_graphs/scripts/country_dashboard.py
miskosz: Good idea, I'll do it. In a short next PR :)

covid_graphs/covid_graphs/scripts/server.py
lukipuki: Sounds good

PR #72, opened 2020-04-28
Event creation date.

miskosz:
Introduced `creation_date` so that we don't need to arbitrarily add one day. Also changes sorting countries in the dropdown by displayed label.

### DIFF
--- a/covid_graphs/covid_graphs/predictions.py
+++ b/covid_graphs/covid_graphs/predictions.py
@@ -9,7 +9,9 @@ from .formula import AtgFormula, Formula
 @dataclass(frozen=True, eq=True)
 class PredictionEvent:
     name: str
+    # Data until this date (inclusive) was used for the prediction.
     date: datetime.date
+    creation_date: datetime.date
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


miskosz:
@rejdi I touched `Rest` interface, would be great if you could have a look.

covid_graphs/covid_graphs/scripts/rest.py
```diff
@@ -72,6 +72,7 @@ def _create_predictions(
                         "max_value": trace.max_value,
                         "date_name": prediction.prediction_event.name,
                         "date": prediction.prediction_event.date,
+                        "creation_date": prediction.prediction_event.creation_date,
                     }
```
rejdi: Addition to the interface is no problem. I'm good to go.

covid_graphs/covid_graphs/predictions.py
```diff
@@ -9,7 +9,9 @@
 @dataclass(frozen=True, eq=True)
 class PredictionEvent:
     name: str
+    # Data until this date (inclusive) was used for the prediction.
     date: datetime.date
+    creation_date: datetime.date
```
rejdi: The comment saves the day for not very usefull name. I'd consider something like `date` -> `last_data_date` and `creation_date` -> `prediction_date` or simply `date` as it's already part of PredictionEvent class. Until now, I was thinking that the old `date` was date of prediction and not last data date used for prediction :smile:

lukipuki (approved):

covid_graphs/covid_graphs/predictions.py
lukipuki: Agree here

miskosz: `last_data_date` is a great naming suggestion. I'll make the change (also creation_date -> prediction_date).

PR #76, opened 2020-05-01
Nginx in front of python server

### LINKED ISSUE #58: Do not run Flask as a production server

rejdi:
#58 Added nginx in front of python server with future support for static files.


docker-compose.yml
```diff
@@ -1,13 +1,23 @@
...
     volumes:
       - "./data:/covid19/data:ro"
-    working_dir: "/covid19"
-    ports:
-      - 127.0.0.1:8081:8081
+    environment:
+      DATA_PATH: "/covid19/data/"
     restart: always
-    command: "covid_graphs.run_server --data-dir data"
+    command: "uwsgi --socket 0.0.0.0:5000 -M --die-on-term -w covid_graphs.wsgi:app"
```
lukipuki: What's the advantage of this approach over FastCGI? https://flask.palletsprojects.com/en/1.1.x/deploying/fastcgi/

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -37,42 +37,54 @@ def run_server(data_dir: Path) -> None:
...
+                p = Process(target=_run_flask_server, kwargs=dict(server=server))
                 p.start()
                 print("Server restarted.")
             sleep(10)
     else:
-        _run_flask_server(server=server, data_dir=data_dir)
+        _run_flask_server(server=server)
+
+
+def uswgi_server(data_dir: Path) -> Flask:
```
lukipuki: This function happens to be used by `uswgi`, but that doesn't mean that we need to call it that.

I think this file should be factored differently.
```suggestion
def setup_server(data_dir: Path) -> Flask:
server = Flask(__name__, template_folder=str(CURRENT_DIR))
... setting up all @server.route's ...
return server
```

That is, change `_setup_server` into a public function.

covid_graphs/nginx/nginx.conf
```diff
@@ -0,0 +1,66 @@
...
+        location / {
+            try_files $uri @proxy;
+        }
+
+        location @proxy {
+            include uwsgi_params;
+            uwsgi_pass dash_server:5000;
+        }
+    }
+}
```
lukipuki: This config seems overly complicated for a server that sits in front of another Nginx -- see the drawing of how we would actually run it: Untitled drawing

I can copy some of these setting to the main Nginx if they are needed for the outside world.

covid_graphs/setup.py
```diff
@@ -18,6 +18,7 @@
         "plotly",
         "pytest",
         "scipy",
+        "uwsgi"
```
lukipuki: I don't think that the package actually uses `uswgi` anywhere. Also, it shouldn't.

covid_graphs/covid_graphs/wsgi.py
```diff
@@ -0,0 +1,10 @@
+from os import getenv
+from pathlib import Path
+
+from covid_graphs.scripts.server import uswgi_server
+
+data_path = getenv(key="DATA_PATH", default="data")
+app = uswgi_server(Path(data_path))
+
+if __name__ == "__main__":
+    app.run()
```
lukipuki: The files associated with the web should be moved elsewhere, e.g. under `/web` in the root of the repository. Also paging @miskosz for an opinion.

I think our `covid_graphs` Python package should end at providing Plotly graphs.  Something else would wrap Plotly graphs inside Dash and Flask. WSGI would run it. The Dash/Flask/WSGI parts should be inside the `/web` directory in another Python package.

docker-compose.yml
```diff
@@ -1,13 +1,23 @@
 version: "2.2"
 
 services:
+  nginx:
```
lukipuki, line 4: As I said before, we're really low on memory on the server, so I'm worried with another Docker image.

- What's the advantage of having a separate Docker nginx image?

- Could we instead run Nginx inside dash_server? Would this save memory?

docker-compose.yml
```diff
@@ -1,13 +1,23 @@
...
   dash_server:
     build: covid_graphs/
     image: covid19_predictions
     volumes:
       - "./data:/covid19/data:ro"
-    working_dir: "/covid19"
-    ports:
-      - 127.0.0.1:8081:8081
+    environment:
+      DATA_PATH: "/covid19/data/"
```
lukipuki, line 18: How could this work when you aren't exposing the port 5000?

covid_graphs/covid_graphs/scripts/server.py
```diff
@@ -37,42 +37,54 @@ def run_server(data_dir: Path) -> None:
...
+
+    @server.route("/covid19/normal/")
+    def covid19_normal_redirect():
+        return redirect(url_for("covid19_all_predictions"))
+
+    @server.route("/covid19/predictions/")
+    def covid19_predictions_redirect():
+        return redirect(url_for("covid19_single_predictions"))
+
     _create_prediction_apps(server=server, data_dir=data_dir)
```
lukipuki, line 72: Why did you take these out of `_run_flask_server`? I think it breaks our data reloading logic.

lukipuki (changes requested):
I think this should be structured differently, both regarding file structure and Docker organization.

covid_graphs/setup.py
rejdi: uwsgi package is standard uwsgi wrapper with python support. On one side, it talk to webserver (nginx) using wsgi protocol, on the other side it translates all calls to python calls using flask.

There is more straighforward option. Just use flask, let it listen on some port using http protocol and forward every call tu it from nginx. I used uWsgi because it was mentioned in #58.

lukipuki: You're missing the point. This is a list of Python dependencies. This package doesn't use `uwsgi` - it's not dependent on it.

rejdi: So proper way is to install it independantly during setup in Dockerfile. Right?

lukipuki: Yes

rejdi: Dependency removed.

docker-compose.yml
rejdi: containers from one docker-compose can talk to each other implicitly

web/README.md
```diff
@@ -0,0 +1,32 @@
...
+covid_web.run_server --data-dir ../data
+```
+Run `covid_web.run_server --help` for help on command parameters.
+
+
+## Local development
+
+This script creates a virtual environment with all development dependencies.
+```sh
+./create_dev_venv.sh
```
lukipuki, line 30: @miskosz What should we do about these scripts for this new web package?

covid_graphs/covid_graphs/scripts/server.py
rejdi: Refactored in new commits.

covid_graphs/covid_graphs/scripts/server.py
rejdi: refactored in new commits.

covid_graphs/covid_graphs/wsgi.py
rejdi: Rearrenged with your help.

docker-compose.yml
rejdi: Proposed fastcgi seems to be just another wrapper over wsgi.

rejdi (commented):
Most suggestions applied.

web/nginx/nginx.conf
```diff
@@ -0,0 +1,66 @@
+worker_processes auto;
```
lukipuki: This config seems overly complicated for a server that sits behind another Nginx, which has most of these settings applied.

web/covid_web/server.py
```diff
@@ -69,10 +57,33 @@ def covid19_predictions_redirect():
...
+    @server.route("/covid19/predictions/")
+    def covid19_predictions_redirect():
+        return redirect(url_for("covid19_single_predictions"))
+
     _create_prediction_apps(server=server, data_dir=data_dir)
     _create_simulation_apps(server=server, data_dir=data_dir)
     _create_rest(server=server, data_dir=data_dir)
+
+
+def uswgi_server(data_dir: Path) -> Flask:
```
lukipuki: This function happens to be used by `uswgi`, but that doesn't mean that we need to call it that.

I think this file should be factored differently.
```suggestion
def setup_server(data_dir: Path) -> Flask:
server = Flask(__name__, template_folder=str(CURRENT_DIR))
... setting up all @server.route's ...
return server
```

That is, change `_setup_server` into a public function.

rejdi: Simplified.

covid_graphs/nginx/nginx.conf
rejdi: Config was simplified.

web/nginx/nginx.conf
rejdi: Simplified in new commit.

lukipuki (approved):

web/README.md
miskosz: Preferably we should have one virtual enviromnment that installs both `covid_graps` and `web` packages.

Issue #61, opened 2020-04-26
Create UI for automatized predictions


lukipuki:

miskosz:
I have created #71 which is the most straightforward way to display the predictions - for a country, display all predictions in one graph. I have left a comment there, I think we should merge only if we see this as a temporary state an it will not break the server :)

miskosz:
Regarding labels, what did you mean?

![80426579-4df55200-88e6-11ea-9203-df40df700841](https://user-images.githubusercontent.com/2885431/80536014-9faae480-8999-11ea-905a-acd6890a2e52.png)

Specifying t0 is tricky.

**Option 1**
I don't really like TeX in the label, but I sense that this might sound controversial. I would prefer if the labels were

- Bodova & Kollar Mar 29
- Bodova & Kollar Apr 12
- Daily prediction Apr 28

and then put the specific values for A, Tg, exp and T0 under the graph.

**Option 2**
With TeX, but values only:
- Mar 29: a=2, tg=3, exp=4, t0=may 10
- Apr 12: ...
- Apr 28*: ...

**Option 3**
Some version of what we have now.

lukipuki:
The big argument against TeX is that it's really slow to render. Also, it gets messy when there're so many numbers shown.

Regarding option three:

- Bodova & Kollar Mar 29 (α=6.23)
- Bodova & Kollar Mar 29 (α=6.23, TG=7.2)

(Or something similar, only give a glimpse of the real parameters and then list all of them below or above. For example the parameter A is not interesting, t0 probably too. )

rejdi:
TeX - both problems can be fixed.
Slowness: we may pre-render formulas as HTML or as Images
Mess: there is no need to show those formulas always. We may show them on request (hover, some click, ...). Another approach is to move the legend under or beside graph.

lukipuki:
@rejdi we shouldn't clutter it with too many formulas. When I plot many predictions, the only parameters I look at are α and TG. The exact formula can be written elsewhere. But I agree that if we can show it e.g. on mouse hover, that would be great (though that sucks on mobile).

lukipuki:
We also want a UI with a slider - the slider picks a date of the automatic prediction. Should we have a separate issue for it? It seems like this one is aimed more for showing more predictions in one graph, while the slider would allow picking just one from a certain date.

rejdi:
@lukipuki No slider yet.

miskosz:
[...]

PR #78, opened 2020-05-02
New web

rejdi:
#61 New web based on react.


lukipuki (changes requested):
Please do a rebase/merge with master first.

rejdi:
@lukipuki I already did.

lukipuki:
Okay, I'm used to looking at individual commits and saw 0fb7132d51abf74110aa95f6a8417357d20eae1d which has been already merged. Also the commit 9ab7346aa45b081f5b981e79832c0fdf589214d1 is included, which has the old long Nginx settings.

It would be nicer if these would be removed or "squashed away". Makes the job of the reviewers easier!

Anyway, this seems like a massive PR (that's why I wanted to view individual commits). I think it should be split, but I don't have good suggestions, as I probably won't review the Javascript parts anyway. Paging @michalburger1 .

miskosz:
@rejdi Can you make a PR which creates an empty React app (that will get changed into what we need in later PRs)? That might help in identifying what is the important code. Or maybe setting up an empty node project and CI setup can be also separated out (I'm not sure since I don't understand details, maybe you can consider.)

PR #80, opened 2020-05-02
Revamp graph labels.

miskosz:
Changelog:

- Revamped labels, removed TeX.
- Colors consistent for the same events.
- CountryGraph remembers correspondence between event and its trace.

### DIFF
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -10,7 +10,7 @@ from plotly.graph_objs import Figure, Layout, Scatter
 
 from .country_report import CountryReport, create_report
 from .formula import FittedFormula, TraceGenerator
-from .predictions import CountryPrediction, PredictionEvent, prediction_db, BK_20200329, BK_20200411
+from .predictions import BK_20200329, BK_20200411, CountryPrediction, PredictionEvent, prediction_db
 
 EXTENSION_PERIOD = datetime.timedelta(days=7)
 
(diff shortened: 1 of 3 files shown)


miskosz:
Screenshot 2020-05-02 at 13 03 41

I used matplotlib tableau color palette with greens for daily predictions. Since I use opacity for various predictions I chose green because humans are most sensitive to changes in green.

miskosz:
Screenshot 2020-05-02 at 13 24 33

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -92,18 +97,32 @@ def adjust_xlabel(date: datetime.date):
...
                 return (date - self.log_xaxis_date_since).days
 
-        colors = ["SteelBlue", "Purple", "Green", "Orange", "Gray"][: len(self.traces)]
+        # Matplotlib colors.
+        color_by_event = {
+            BK_20200329: "rgb(255, 123, 37)",
+            BK_20200411: "rgb(0, 121, 177)",
+        }
+        extra_colors = [
+            "rgba(43, 161, 59, 0.4)",
```
lukipuki, line 106: This color is too close to our light green background, we should probably increase opacity.

lukipuki (approved):

covid_graphs/covid_graphs/country_graph.py
miskosz: We have to change the light green background, it does not make sense with multiple predictions.

Since we don't use the color, I'd leave that for @rejdi, the colors will adjust with the new frontend anyway I believe.

lukipuki: I don't think it needs to be dropped entirely. The shade of green can change, each shaded area will correspond to another prediction date.

PR #83, opened 2020-05-02
Add prediction_generator.py for generating predictions

### LINKED ISSUE #74: Store automated predictions offline
Our automated predictions should be generated every night when new data is downloaded.

Main storage format of predictions should be protobuf and converted to JSON for the web app.

lukipuki:
I've also added a proto to store the fitted formulas and a method to
store them.


covid_graphs/covid_graphs/formula.py
```diff
@@ -117,20 +117,23 @@ def get_trace_generator(self, country_report: CountryReport) -> TraceGenerator:
         # The choice of date zero is in theory arbitrary.
         date_zero = country_report.dates[0]
         xs = [(date - date_zero).days for date in country_report.dates[: until_idx + 1]]
-        fit = fit_atg_model.fit_atg_model(
+        self.fit = fit_atg_model.fit_atg_model(
```
lukipuki, line 123: I've changed `fit` into `self.fit`, which is later accessed in `store_predictions`. This is ugly, but I'm okay with taking some technical debt for now. It's ugly, because `get_trace_generator` has to be called first before accessing `self.fit`.

If you have simple suggestions to make this better, I can incorporate them now. Otherwise it will have to wait after our release.

One alternative is to move `self.fit` calculation into the constructor, but the constructor would need to take additional parameter `CountryReport`.

lukipuki: Okay, went with the constructor. It was possible at both call-sites.

covid_graphs/covid_graphs/formula.py
```diff
@@ -108,31 +109,37 @@ def formula(x: float):
 
 @dataclass
```
miskosz: You can remove the decorator. It was useful only for automatic init method.

covid_graphs/covid_graphs/formula.py
miskosz: Since `Formula`/`Trace` are very generic classes (probably because of early experiments with Slovakia) we are missing now a general class for both B&K and fitted predictions. My hunch is that this is the reason for the ugliness.

We can definitely make/repurpose one, however a different PR might be more suitable.

covid_graphs/covid_graphs/formula.py
```diff
@@ -108,31 +109,37 @@ def formula(x: float):
...
-
         # Counterintuitively, `date` + `timedelta` results in `date`.
-        whole_day_offset = np.floor(fit.t0)
-        start_date = date_zero + datetime.timedelta(days=whole_day_offset)
-        display_at_least_until = _get_display_at_least_until(
-            tg=fit.tg, exp=fit.exp, start_date=start_date,
+        whole_day_offset = np.floor(self.fit.t0)
+        self.fit.shift_forward(whole_day_offset)
+
+        date_zero = country_report.dates[0]
```
miskosz: Duplicate line, see L122.

covid_graphs/covid_graphs/formula.py
```diff
@@ -108,31 +109,37 @@ def formula(x: float):
...
 
+    def get_trace_generator(self, country_report: CountryReport) -> TraceGenerator:
+        label = _create_atg_label("Daily prediction", tg=self.fit.tg, alpha=self.fit.exp)
         return TraceGenerator(
-            func=lambda x: fit.predict(x + whole_day_offset),
-            start_date=start_date,
-            display_at_least_until=display_at_least_until,
+            func=self.fit.predict,
+            start_date=self.start_date,
+            display_at_least_until=self.display_at_least_until,
```
miskosz: Why don't you create the trace object in the `init` and just return it here?

covid_graphs/covid_graphs/pb/atg_prediction.proto
```diff
@@ -0,0 +1,23 @@
...
+}
+
+message AtgTraceGenerator {
+  Date last_data_date = 1;
+
+  double alpha = 5;
+  double tg = 6;
+  double a = 7;
+  double t0 = 8;
+  Date start_date = 9;
```
miskosz, line 18: Please add a comment on the meaning/relation of `t0` and `start_date` (it is not clear to me).

covid_graphs/covid_graphs/pb/atg_prediction.proto
```diff
@@ -0,0 +1,23 @@
...
+}
+
+message AtgTraceGenerators {
```
miskosz: Minor: `CountryAtgTraceGenerators`?

covid_graphs/covid_graphs/prediction_generator.py
```diff
@@ -0,0 +1,75 @@
...
+
+@click.command(help="COVID-19 country predictions calculation")
+@click.argument(
+    "data_dir", required=True, type=click_pathlib.Path(exists=True),
+)
+@click.argument(
+    "short_country_name", required=True, type=str,
+)
+@click.argument(
+    "prediction_dir", required=True, type=click_pathlib.Path(exists=True),
```
miskosz: I'd call it `output_dir`. Also, it would be nice to add help messages :)

miskosz: You can consider removing `exists=True` and adding to the function:
```
prediction_dir.mkdir(parents=True, exist_ok=True)
```

covid_graphs/covid_graphs/prediction_generator.py
```diff
@@ -0,0 +1,75 @@
...
+        CountryPrediction(
+            prediction_event=PredictionEvent(
+                name=f"daily_fit_{last_data_date.strftime('%Y_%m_%d')}",
+                last_data_date=last_data_date,
+                prediction_date=last_data_date,
+            ),
+            country=report.short_name,
+            formula=FittedFormula(until_date=last_data_date, country_report=report),
+        )
+        for last_data_date in dates
```
miskosz, line 30: Can we create non-sensical predictions here? Maybe we should add:
```py
if last_data_date in report.dates
```

covid_graphs/covid_graphs/prediction_generator.py
```diff
@@ -0,0 +1,75 @@
...
+)
+def store_predictions(data_dir: Path, short_country_name: str, prediction_dir: Path) -> None:
+    country_report = create_report(
+        data_dir / f"{short_country_name}.data", short_name=short_country_name
+    )
+    last_date_dates = country_report.dates[-PREDICTION_DAYS:]
+    fitted_formulas = list(
+        FittedFormula(until_date=last_data_date, country_report=country_report)
+        for last_data_date in last_date_dates
+    )
```
miskosz: Nit: `list(...)` -> `[...]`.

miskosz: Also I find it better to explicitly tie the formulas to dates (dicts are fast in python):
```
fitted_formula_by_date = {
    last_data_date: FittedFormula ...
}
```
Then you can remove `zip` below and use `fitted_formula_by_date.items()`.

covid_graphs/covid_graphs/prediction_generator.py
```diff
@@ -0,0 +1,75 @@
...
+
+    trace_generators = AtgTraceGenerators()
+    trace_generators.short_country_name = short_country_name
+    trace_generators.long_country_name = country_report.long_name
+    for last_date_date, formula in zip(last_date_dates, fitted_formulas):
```
miskosz: typo: `last_datA_date `

miskosz: also on L48

covid_graphs/covid_graphs/prediction_generator.py
```diff
@@ -0,0 +1,75 @@
...
+        generator = AtgTraceGenerator()
+        generator.last_data_date.year = last_date_date.year
+        generator.last_data_date.month = last_date_date.month
+        generator.last_data_date.day = last_date_date.day
+
+        trace_generator = formula.get_trace_generator(country_report)
```
miskosz, line 66: Note: Regarding the "elegance" comment with `self.fit`. I see now why did you do it. I think a future solution could be that current `TraceGenerators` return the proto.

covid_graphs/covid_graphs/pb/atg_prediction.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+message Date {
+  uint32 day = 1;
+  uint32 month = 2;
+  uint32 year = 3;
+}
+
+message AtgTraceGenerator {
```
miskosz: I feel quite strongly that this should not be called a "generator", and especially because this is an interface we should choose a good name. I suggest something along `AtgParameters` or `AtgModelParameters` (I'm fine with `...Params` as well).

covid_graphs/setup.py
```diff
@@ -24,10 +24,11 @@
     packages=find_packages(),
     entry_points={
         "console_scripts": [
+            "covid_graphs.prepare_data = covid_graphs.prepare_data:main",
             "covid_graphs.show_country_plot = covid_graphs.country_graph:show_country_plot",
             "covid_graphs.show_heat_map = covid_graphs.heat_map:show_heat_map",
             "covid_graphs.show_scatter_plot = covid_graphs.scatter_plot:show_scatter_plot",
-            "covid_graphs.prepare_data = covid_graphs.prepare_data:main",
+            "covid_graphs.store_predictions = covid_graphs.prediction_generator:store_predictions",
```
miskosz: Minor: I feel that `generate_predictions` describes better what it does.

miskosz (commented):
Looks good! I have a few comments. I would normally like to have another look before approval but I don't want to block you (will not look at it tomorrow before noon), so I am not formally requesting changes (well it would not block merging but anyway 😛)

covid_graphs/covid_graphs/fit_atg_model.py
```diff
@@ -16,6 +16,9 @@ def predict(self, x: float) -> float:
         ys = _model(params=[self.a, self.tg, self.exp, self.t0], xs=np.array([x]))
         return ys[0]
 
+    def shift_forward(self, day_count: int) -> None:
```
miskosz, line 20: This feels hacky.

covid_graphs/covid_graphs/formula.py

covid_graphs/covid_graphs/formula.py

covid_graphs/covid_graphs/prediction_generator.py
lukipuki: Done

covid_graphs/covid_graphs/prediction_generator.py
lukipuki: Done

covid_graphs/setup.py
lukipuki: Done

covid_graphs/covid_graphs/prediction_generator.py
lukipuki: Yeah, that's probably the right place to do it. Will do it next time we're in need of refactoring/simplifying :)

covid_graphs/covid_graphs/prediction_generator.py
lukipuki: Done (though without a help message, I thought these are hard to do with these positional arguments?)

covid_graphs/covid_graphs/pb/atg_prediction.proto
lukipuki: `AtgParameters` is good.

covid_graphs/covid_graphs/formula.py
lukipuki: I thought about it but then the function would be so short :grin: I've changed it.

covid_graphs/covid_graphs/pb/atg_prediction.proto
lukipuki: `CountryAtgParameters`

lukipuki (commented):
I've addressed most issues, I'll get back to the remaining issues regarding `t0` tomorrow.

covid_graphs/covid_graphs/pb/atg_prediction.proto
lukipuki: I've changed t0 to `offset` and added a comment.

covid_graphs/covid_graphs/fit_atg_model.py
lukipuki: Because the usage of `AtgModelFit` sucks! :smile: I think the implementation details (`t0`) leak and make the usage of `AtgModelFit` unnecessarily complicated.

To use the `predict` function, we have to do `func=lambda x: fit.predict(x + whole_day_offset)`! The funny thing is that the implementation immediately subtracts it away: `x + whole_day_offset - t0`. My change simplifies this non-sense, so that we only do `x - t0`.

However, I agree that that's hacky, but I don't have time to fix this file.

`t0` shouldn't be part of the final usage via `predict`. In the new proto I have start date and a variable called `offset`, which is less than 1 and together with start date marks the exact start of the curve. Basically start_date and offset together form `t0`.

Perhaps this file can get inspiration from it, but at best I can add a TODO now.

covid_graphs/covid_graphs/prediction_generator.py
lukipuki: Done

PR #85, opened 2020-05-03
Introduce slider UI element

### LINKED ISSUE #79: Create UI for choosing the date of predictions
We'd like to have a UI similar to the one already running at https://graphs.lukipuki.sk/covid19/predictions/single

Instead of choosing the date of prediction in a dropdown, there will be a slider moving the green zone.

This could be implemented in pure Plotly or in JS. I'll try to do the Plotly version this weekend.

lukipuki:
* I've renamed the old GraphType to GraphAxisType
* The new GraphType is either slider, single prediction or multiple
  predictions.

The startup of the server is really slow, because it calculates about
300 predictions. This will get better with static predictions stored on
the disk.

### DIFF
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -11,12 +11,21 @@ from plotly.graph_objs import Figure, Layout, Scatter
 from .country_report import CountryReport, create_report
 from .formula import TraceGenerator
 from .prediction_generator import get_fitted_predictions
-from .predictions import BK_20200329, BK_20200411, CountryPrediction, prediction_db
+from .predictions import BK_20200329, BK_20200411, CountryPrediction
 
 EXTENSION_PERIOD = datetime.timedelta(days=7)
 
 
 class GraphType(Enum):
+    Slider = "slider"
+    SinglePrediction = "single"
+    MultiPredictions = "multi"
+
+    def __str__(self):
+        return self.value
+
+
+class GraphAxisType(Enum):
     Linear = "linear"
     SemiLog = "semi-log"
     LogLog = "log-log"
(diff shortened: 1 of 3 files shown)


covid_graphs/covid_graphs/country_graph.py
```diff
@@ -11,12 +11,21 @@
...
 class GraphType(Enum):
+    Slider = "slider"
+    SinglePrediction = "single"
+    MultiPredictions = "multi"
+
+    def __str__(self):
+        return self.value
+
+
+class GraphAxisType(Enum):
```
miskosz, line 28: +1

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -80,22 +89,54 @@ def __init__(
...
             max(trace.max_value for trace in self.trace_by_event.values()),
         )
         self.log_yrange = [
             math.log10(max(1, self.cropped_cumulative_active.min())) - 0.3,
-            math.log10(max_value) + 0.3,
+            math.log10(self.max_value) + 0.3,
+        ]
+
+    def _create_slider(self, traces: List[Scatter]):
+        SUBTRACE_COUNT = 4
```
miskosz: Minor: All caps are usually for global constants. In this instance, this could be a good solution:
`_create_slider(self, traces: List[Scatter], subtrace_count: int = 4) -> Dict[str, Any]: `

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -80,22 +89,54 @@ def __init__(
...
+
+        last_position = len(self.trace_by_event) - 1
+        # Mark last trace as visible
+        for i in range(last_position * SUBTRACE_COUNT, (last_position + 1) * SUBTRACE_COUNT):
+            traces[i].visible = True
```
miskosz: Nit:
```
for trace in traces[last_position * SUBTRACE_COUNT : (last_position + 1) * SUBTRACE_COUNT]:
    trace.visible = True
```

covid_graphs/covid_graphs/country_graph.py
miskosz: Oh I see what this constant is... number of traces per legend item. Maybe it can indeed be moved to the global scope at the top of the file.

I also wonder if the traces can be selected based on `legendgroup` attribute instead.

covid_graphs/covid_graphs/country_graph.py
miskosz: Actually, maybe even this?
```
for trace in traces[-SUBTRACE_COUNT:]:
    trace.visible = True
```

web/covid_web/country_dashboard.py
```diff
@@ -178,42 +170,27 @@ def _create_buttons(
...
+                    dates=self.report_by_short_name[country_short_name].dates[-14:],
+                ),
+            )
+            for country_short_name in prediction_db.get_countries()
+        }
 
         @app.callback(
             Output("country-graph", component_property="figure"),
-            [
-                Input("prediction-event", "value"),
```
miskosz, line 199: So we allow sliding through the two BK predictions now? I am a bit confused now how `SingleCountry` page functionality is changed. Also do we now mix BK and Daily predictions on the daily predictions page, or BK is always visible?

miskosz (approved):
LGTM.

covid_graphs/covid_graphs/country_graph.py
lukipuki: I went with `legendgroup`, which simplifies the logic and is less error-prone.

covid_graphs/covid_graphs/country_graph.py
lukipuki: Not relevant anymore, filtering based on `legendgroup` now.

web/covid_web/country_dashboard.py
lukipuki: Single country prediction is for automatic predictions only. I've mentioned this on Slack - to keep the number of web pages and dashboard manageable, we have to drop functionality from time to time.

BK predictions are still visible on the `/covid19/all` page, which is the most popular one anyway :smile:

PR #86, opened 2020-05-03
Calculate predictions after downloading the data

lukipuki:


data/download_all.zsh
```diff
@@ -15,9 +15,12 @@ countries=("United Kingdom" "South Korea" Austria Iceland Jordan Switzerland Cro
...
 do
   if [[ -z $short_names["${country}"] ]]; then
-    covid_graphs.prepare_data ${(p)country}
+    covid_graphs.prepare_data ${country}
+    covid_graphs.generate_predictions . ${country} predictions
   else
-    covid_graphs.prepare_data ${(p)country} --short_name $short_names["${country}"]
+    short_name=$short_names["${country}"]
+    covid_graphs.prepare_data ${(p)country} --short_name ${short_name}
+    covid_graphs.generate_predictions . ${short_name} predictions
```
rejdi: It's a bit odd that `prepare_data` implicitly expect data in current dir and `generate_predictions` require explicit input dir. It's not problem of this script, but rather an inconsistences of python scripts. Suggested change would be that `generate_predictions` took input file and output file, with no need to directory.

lukipuki: I think you have certain things switched up, e.g. `prepare_data` doesn't read any files. But the interface of `generate_predictions`, where we have a directory as well as country name is unnecessary. Country name can be read from the proto.

Storing the short country name inside the proto ended up being quite substantial change, but should help us in the future.

rejdi (approved):

PR #87, opened 2020-05-03
New web base

rejdi:
As requested by @miskosz this is just a base for new web. ref: #78

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,6 +18,12 @@ services:
       DATA_PATH: "/covid19/data/"
     restart: always
     command: "uwsgi --uid www-data --gid www-data --socket 0.0.0.0:5000 --die-on-term -w covid_web.wsgi:app"
+  react-web:
+    build: web/react-web/
+    image: react-web
+    volumes:
+      - "react-web:/react-web"
+    command: "bash -c ' rm -rf /react-web/* && cp -a . /react-web/'"
 
 volumes:
   react-web:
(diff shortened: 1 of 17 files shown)


docker-compose.yml
```diff
@@ -18,6 +18,12 @@ services:
       DATA_PATH: "/covid19/data/"
     restart: always
     command: "uwsgi --uid www-data --gid www-data --socket 0.0.0.0:5000 --die-on-term -w covid_web.wsgi:app"
+  react-web:
+    build: web/react-web/
+    image: react-web
+    volumes:
+      - "react-web:/react-web"
+    command: "bash -c ' rm -rf /react-web/* && cp -a . /react-web/'"
```
lukipuki, line 26: What does this line do?

web/react-web/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM node:latest
+
+COPY . /app
+WORKDIR /app
+RUN npm install
+RUN npm run build
+RUN mv build /
+WORKDIR /build
+RUN rm -rf /app
```
lukipuki: If I get what's going on, these might be better names:
```suggestion
COPY . /app-source
WORKDIR /app-source
RUN npm install
RUN npm run build
RUN mv build /app
WORKDIR /app
RUN rm -rf /app-source
```

lukipuki (approved):
I approve the non-JS part

docker-compose.yml
rejdi: Dockerfile from react-web just build the web and make static files that resides in image. This line move them to shared volume called `react-web` and docker container `react-web` ends. Shared volume is then accessible by nginx and web can be access at http://localhost:8081/covid19/graphs.

lukipuki: Okay, than this container should be called `react-web-builder` instead.

rejdi: Applied

PR #89, opened 2020-05-03
New react-based web

rejdi:
#61 New web based on react and updated graph styling. No slider yet.


web/react-web/src/App.test.js
```diff
@@ -4,6 +4,6 @@ import App from './App';
 
 test('renders learn react link', () => {
   const { getByText } = render(<App />);
-  const linkElement = getByText(/Sample app/i);
+  const linkElement = getByText(/April 13 predictions: Personal communication/i);
```
lukipuki: It's April 12

web/react-web/src/Sidebar/index.js
```diff
@@ -0,0 +1,113 @@
...
+
+                            return (
+                                <>
+                                    <div className='buttons'>
+                                        <button className={activeTab === 0 ? 'active' : ''} onClick={this.selectTab(0)}>About</button>
+                                        <button className={activeTab === 1 ? 'active' : ''} onClick={this.selectTab(1)}>Results</button>
+                                    </div>
+                                    {activeTab === 0 &&
+                                        <>
+                                            <h1>COVID-19 predictions of Boďová and Kollár</h1>
```
lukipuki, line 51: Please add a TODO that this text should be generated from Markdown that's shared with Python and we will fix it in another PR. Actually, we might need to serve it from Python via REST, as the text might be modified per page.

web/react-web/src/assets/Hamburger_icon.svg
```diff
@@ -0,0 +1 @@
+<?xml version="1.0" ?><!DOCTYPE svg  PUBLIC '-//W3C//DTD SVG 1.1//EN'  'http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd'><svg xmlns:xlink="http://www.w3.org/1999/xlink" height="32px" id="Layer_1" style="enable-background:new 0 0 32 32;" version="1.1" viewBox="0 0 32 32" width="32px" xml:space="preserve" xmlns="http://www.w3.org/2000/svg"><path d="M4,10h24c1.104,0,2-0.896,2-2s-0.896-2-2-2H4C2.896,6,2,6.896,2,8S2.896,10,4,10z M28,14H4c-1.104,0-2,0.896-2,2  s0.896,2,2,2h24c1.104,0,2-0.896,2-2S29.104,14,28,14z M28,22H4c-1.104,0-2,0.896-2,2s0.896,2,2,2h24c1.104,0,2-0.896,2-2  S29.104,22,28,22z"/></svg>
```
lukipuki, line 1: Why Hamburger?

lukipuki (approved):
Again, for the non-JS part.

web/react-web/src/App.test.js
rejdi: Interesting, this was copy&pasted from the past. Changed.

web/react-web/src/Sidebar/index.js
rejdi: Added.

web/react-web/src/assets/Hamburger_icon.svg
rejdi: It's used to be a name for it. I don't know where it comes from. https://www.google.com/search?q=hamburger+menu+icon

web/react-web/README.md
```diff
@@ -0,0 +1,29 @@
+## Development
+### Local development
+You need to have [`NodeJS`](https://nodejs.org/) with [`npm`](https://www.npmjs.com/) or [`yarn`](https://yarnpkg.com/) and backend (see [README](../covid_graphs/README.md)).
```
miskosz, line 3: Me as a non-expert thought that "`yarn` is a newer better `npm`", so I was wondering if we should just use `yarn` in the readme.

web/react-web/README.md
```diff
@@ -0,0 +1,29 @@
...
+1. run `npm install` or `yarn install` if not already did.
+1. run `npm build` or `yarn build`.
+
+Final application should be in `build` folder
+
+## Running the whole server
+
+In the root directory of the repository, run `docker-compose`.
+```sh
+docker-compose up --build # Optionally add -d for deamon
```
miskosz: typo: daemon (I know it was copied...)

web/react-web/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+You can then access the server locally at [localhost:8081](http://127.0.0.1:8081/covid19/graphs/).
+
+Application expects that after deployment, following paths will be answered by backend.
+* `/covid19/rest/data/*`
+* `/covid19/rest/predictions/list/`
+* `/covid19/rest/predictions/data/*`
+
```
miskosz, line 29: Shall we get rid of `README.react.md` mow?

web/react-web/src/App.css
```diff
@@ -1,3 +1,129 @@
...
+  }
+
+  .App .sidebar {
+    max-width: initial;
+    min-width: initial;
+    position: absolute;
+    min-height: 100%;
+    overflow: initial;
+  }
+}
```
miskosz, line 129: Does this CSS come from a template, or was it hand-written? If the latter, do we have any concerns (or aspirations :)) of it working well across different devices?

web/react-web/src/App.js
```diff
@@ -1,12 +1,125 @@
...
+                }
+            })
+            .then((data) => {
+                const newData = {...this.state.dataContext};
+                newData[what] = data;
+                this.setState({
+                    dataContext: newData
+                });
+            })
+            .catch(() => {});
```
miskosz: Without knowing what's going on, this line is suspicious. If there was a problem shouldn't we handle it somehow? (At least a printout..)

web/react-web/src/App.js
```diff
@@ -1,12 +1,125 @@
...
+        })
+    };
+
+    fetchData = (what) => {
+        fetch(`/covid19/rest/${what}`, {cache: 'no-cache'})
+            .then((response) => {
+                if (response.ok) {
+                    return response.json();
+                } else {
+                    throw new Error();
```
miskosz: Shouldn't we set at least an error message?

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,240 @@
...
+
+const colors = ["#7cb5ec", "#434348", "#90ed7d", "#f7a35c", "#8085e9", "#f15c80", "#e4d354", "#2b908f", "#f45b5b", "#91e8e1"];
+
+const predefinedOptions = {
+    chart: {
+        zoomType: 'x',
+        colors,
+    },
+    title: {
+        text: 'USD to EUR exchange rate over time'
```
miskosz: This does not seem right.

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,240 @@
...
+
+    render() {
+        const {
+            data,
+            options
+        } = this.props;
+
+        const {
+            mode,
+            relative,
```
miskosz: Can we think of more descriptive names for these two variables?

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,240 @@
...
+
+        data.forEach((data, index)=> {
+            let maxXValue = null;
+            let predictionXValue = null;
+            let name = '';
+            const zones = [];
+            let dashStyle = 'line';
+
+            if (data.type === 'prediction') {
+                if (relative) {
```
miskosz: I'm not sure what `relative` means.

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,240 @@
...
+                    maxXValue = data.date_list.indexOf(data.max_value_date) + 1;
+                    predictionXValue = data.date_list.indexOf(data.prediction_date) + 1;
+                } else {
+                    maxXValue = Date.parse(data.max_value_date);
+                    predictionXValue = Date.parse(data.prediction_date);
+                }
+
+                name = data.description.replace('%PREDICTION_DATE%', `<br/>${new Date(Date.parse(data.prediction_date)).toLocaleDateString()}`);
```
miskosz: Oh well, a hack from python leaking to js :) I'd suggest to replace the string on the backend before serving.

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,240 @@
...
+                    x = Date.parse(date);
+                }
+
+                if (x === maxXValue) {
+                    return {
+                        enabled: true,
+                        x,
+                        y,
+                        marker: {
+                            symbol: 'triangle-down',
```
miskosz: Nit: `triangle-up` is maybe resembling a peak more? :)

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,240 @@
...
+
+                let x = null;
+                if (relative) {
+                    x = index + 1;
+                } else {
+                    x = Date.parse(date);
+                }
+
+                if (x === maxXValue) {
+                    return {
```
miskosz: Is it possible to set some variables instead and have only one return in this function? That might prevent unintentional side effects when editing parts of this function.

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,240 @@
...
+
+                return [x, y];
+            });
+
+            series.push({
+                type: 'line',
+                dashStyle,
+                name,
+                marker: {
+                    enabled: data.type !== 'prediction'
```
miskosz, line 166: What does this `enabled` mean everywhere?

web/react-web/src/Graph/Graph.js
miskosz: I understood `mode` from the code below, maybe `axis_type` or similar could be suitable?

miskosz:
I am having a look.

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,240 @@
...
+                finalOptions.yAxis.type = 'logarithmic';
+                finalOptions.xAxis.type = 'linear';
+                break;
+            case MODE_LOG_LOG:
+                finalOptions.xAxis.type = 'logarithmic';
+                finalOptions.yAxis.type = 'logarithmic';
+                break;
+        }
+
+        if (relative) {
```
miskosz: It is not clear to me what this `if-else` block does, and it seems non-trivial. Maybe a comment can be added? Or maybe the functionality can be extracted to a function (or two?).

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,240 @@
...
+            finalOptions.tooltip.formatter = function() {
+                if (this.points) {
```
miskosz: Why do we have data sometimes in `this.points` and sometimes in `this.point` (I assume)? Can it be consolidated?

web/react-web/src/Sidebar/Countries.js
```diff
@@ -0,0 +1,69 @@
...
+        } = this.props;
+
+        if (!predictions.data) {
+            return null;
+        }
+
+        const countries = new Set();
+        const predictionDates = Object.keys(predictions.data).sort();
+        for (let dateKey in predictions.data) {
+            if (predictions.data.hasOwnProperty(dateKey)) {
```
miskosz: Maybe a noob js question, but why do we need to check if `dateKey` is in `data` if we are iterating over keys in `data`? And can't we iterate directly over values?

web/react-web/src/Sidebar/Countries.js
```diff
@@ -0,0 +1,69 @@
...
+        });
+
+        return (
+            <table>
+                <thead>
+                    <tr>
+                        <th rowSpan={2}>Country<br/>(active cases)</th><th colSpan={predictionDates.length}>Predictions</th>
+                    </tr>
+                    <tr>
+                        {predictionDates.map((value, index) => <th key={index}>{new Date(Date.parse(predictions.data[value].label)).toLocaleDateString()}</th>)}
```
miskosz: Is the extra `new Date(..)` wrapper needed around `Date.parse(...)`?

web/react-web/src/Sidebar/index.js
miskosz: I think the text should not be served via REST (anyway we want to deprecate flask). But it is weird that it lives in js, I'm sure there must be a templating solution.

web/react-web/src/Sidebar/index.js
```diff
@@ -0,0 +1,114 @@
...
+                    <GraphDetailContext.Consumer>
+                        {graphDetailContext => {
+                            const dates = [];
+                            const {relative, mode} = graphDetailContext.options;
+                            const {activeTab} = this.state;
+
+                            for (let key in context.data) {
+                                const item = context.data[key];
+                                const date = new Date(Date.parse(item.label));
+                                const className = `tab${this.state.activeDateKey === key ? ' active' : ''}`;
```
miskosz, line 37: Is this referring to a css class?

web/react-web/src/assets/Hamburger_icon.svg
miskosz: Yes, hamburger icon is a standard name :)

miskosz (commented):
I am happy to see that frontend is coming together. I have left a few (well... some) comments. Please bear in mind they come from someone who has not worked with JS for some time, and someone with no experience with React.

I think we'd like different people to be able to contribute to the frontend code, which of course is difficult if we have little experience with the technology. Most of my comments are about how to make it easier for the rest of us to understand the code - I pointed out a few places where the naming was not clear or where the code structure was surprising to me. (Maybe some can be remedied by simply leaving a comment.)

web/react-web/src/Sidebar/Countries.js
miro-balaz: This is the only version for iterating in old javascript. [source](https://zellwk.com/blog/looping-through-js-objects/)  in ES6 it is possible to to write `predictions.values.forEach(x=>{...  }) `

web/react-web/src/Sidebar/Countries.js
miro-balaz: lmgtfu, https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse

As i understand it is the same.

web/react-web/README.md
rejdi: I'm used to use yarn, however as we are using travis a docker for building, npm is better choice for several reasons:
* it's a part of official nodejs docker image, while yarn isn't
* npm more standardized than yarn. Yarn was created to be a faster replacement of npm and it tries to maintain compatibility as much as possible with npm
* unfortunately, yarn is using it's own lock file, so if we'd mix two different package managers, we may get unstable results

web/react-web/src/App.css
rejdi: It was handwritten. Yes, we have both :smile:. It's already suited for mobile. At this stage, I wanted to keep it as simple as possible, because current design is not a rocket science.

web/react-web/README.md
rejdi: Yes.

web/react-web/README.md
rejdi: Fixed.

web/react-web/src/App.js
rejdi: You're right. This one deserves at least some TODO.

web/react-web/src/App.js
rejdi: Right, `response.statusText` gets here.

web/react-web/src/Graph/Graph.js
rejdi: Thanks! Copy-past mistake. This value is replaced on every render.

web/react-web/src/Graph/Graph.js
rejdi: mode -> axesType
relative -> xAxisRelative

web/react-web/src/Graph/Graph.js
rejdi: Relative means - aligned to start. So first value means first day. It allows us to compare countries on each day and tell which "perform" better over time.

web/react-web/src/Graph/Graph.js
rejdi: I was thinking of that too, but I've made it here, because it get replaced by localised date, according to user preference.

web/react-web/src/Graph/Graph.js
rejdi: I was trying to avoid confusion about - Hey! They wanted to get it even higher! :)

web/react-web/src/Graph/Graph.js
rejdi: It is and fixed.

web/react-web/src/Graph/Graph.js
rejdi: This property is related to https://api.highcharts.com/highcharts/series.line.marker. In `render` function, I'm preparing data and options for Highcharts. This particular enabled means, for data serie, render marker on each day. For predictions, just plot line without markers (except on prediction date and culmination day).

web/react-web/src/Graph/Graph.js
rejdi: I moved last line of each block to top as it's descriptive. however it's still a bit complicated. For now, I don't have idea how to make it better, so every suggestion is welcome! I think, fix related to `relative` variable name may help here.

web/react-web/src/Graph/Graph.js
rejdi: I'll put comment here. Tooltip can be shared or per serie. If it's shared (this.points exists), we need to extract data from every serie.
ref: https://api.highcharts.com/highcharts/tooltip.formatter

web/react-web/src/Sidebar/Countries.js
rejdi: > Maybe a noob js question, but why do we need to check if `dateKey` is in `data` if we are iterating over keys in `data`? And can't we iterate directly over values?

Objects could be derived. I that case, we're checking whether it's own property or derived one. This time, this could be ommited, however I was just getting warning from IDE.

> This is the only version for iterating in old javascript. [source](https://zellwk.com/blog/looping-through-js-objects/) in ES6 it is possible to to write `predictions.values.forEach(x=>{... }) `

Right, however in this case this is object. So it'd look like
```
Object.keys(yourobject).forEach((key) => {
...body...
});
```
or
```
Object.entries(predictions.data).forEach(([dateKey, prediction]) => {
  prediction.countries.forEach(country => {
    countries.add(country);
  });
});
```
I'll use later one.

web/react-web/src/Sidebar/Countries.js
rejdi: Right, changed!

web/react-web/src/Sidebar/index.js
rejdi: yes.

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,241 @@
...
+
+const colors = ["#7cb5ec", "#434348", "#90ed7d", "#f7a35c", "#8085e9", "#f15c80", "#e4d354", "#2b908f", "#f45b5b", "#91e8e1"];
+
+const predefinedOptions = {
+    chart: {
+        zoomType: 'x',
+        colors,
+    },
+    title: {
+        text: ''
```
miskosz, line 14: We can set `COVID 19 Chart` or something like that, no? (No idea where it shows up, but somewhere it does I guess) :)

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,241 @@
...
+
+    render() {
+        const {
+            data,
+            options
+        } = this.props;
+
+        const {
+            axesType,
+            xAxisRelative,
```
miskosz: Minor: We can consider `isXAxisRelative` to make it clearer it is a bool.

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,241 @@
...
+                break;
+        }
+
+        if (xAxisRelative) {
+            finalOptions.xAxis.title.text = 'Day since relevant number of cases';
+
+            finalOptions.tooltip.formatter = function() {
+                //https://api.highcharts.com/highcharts/tooltip.formatter
+                if (this.points) {
+                    //shared tooltip, we need extract data from every serie
```
miskosz: Nit: `serie` -> `series` (the word has the same singular and plural)

web/react-web/src/Graph/Graph.js
```diff
@@ -0,0 +1,241 @@
...
+                    return this.points.reduce((s, point) => {
+                        const name = point.series.name;
+                        return `${s}<br/><strong>${name}:</strong> ${Math.round(point.y)}`;
+                    }, `<div style="text-align: center"><strong>Day ${this.x}</strong></div>`);
+                } else {
+                    const name = this.point.series.name;
+                    return `<div style="text-align: center"><strong>Day ${this.x}</strong></div><br/><strong>${name}:</strong> ${Math.round(this.point.y)}`;
```
miskosz: An idea what might make this code less repetitive: Could we define a function
```
xaxis_text(x, xAxisRelative)
```
That returns `Day ${this.x}` or `${date.toLocaleDateString()}`?

miskosz (approved):
LGTM

web/react-web/src/Graph/Graph.js
rejdi: It's on top of chart - main title, but it's changed every time you pick some country. This line is just predefined text. The whole concept is, that I have initial config for highcharts `const predefinedOptions = {...` and during render I change only what's needed. Title is changing.

web/react-web/src/Graph/Graph.js
rejdi: Changed.

web/react-web/src/Graph/Graph.js
rejdi: Fixed.

web/react-web/src/Graph/Graph.js
rejdi: OMG, right! :)

PR #91, opened 2020-05-07
Serialize/deserialize FittedFormula

### LINKED ISSUE #74: Store automated predictions offline

lukipuki:
There are two commits, one regarding serialization, one regarding deserialization. You can view them separately, but it seems it's possible to to review them together.

### DIFF
--- a/covid_graphs/covid_graphs/formula.py
+++ b/covid_graphs/covid_graphs/formula.py
@@ -9,6 +9,7 @@ import numpy as np
 from . import fit_atg_model
 from .country_report import CountryReport
 from .fit_atg_model import AtgModelFit
+from .pb.atg_prediction_pb2 import AtgParameters
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


miro-balaz (commented):
LGTM, but I have no experience with protobuf.

covid_graphs/covid_graphs/prediction_generator.py
```diff
@@ -44,31 +44,16 @@ def generate_predictions(filename: Path, output_dir: Path) -> None:
...
 
     short_country_name = country_report.short_name
-    arg_parameters = CountryAtgParameters()
-    arg_parameters.long_country_name = country_report.long_name
-    arg_parameters.short_country_name = short_country_name
+    country_atg_parameters = CountryAtgParameters()
+    country_atg_parameters.long_country_name = country_report.long_name
+    country_atg_parameters.short_country_name = short_country_name
     for last_data_date, formula in fitted_formulas.items():
-        # TODO(lukas): create a serialization function inside the trace generator
```
miro-balaz: last_data_data is not used in this block

miro-balaz (approved):

covid_graphs/covid_graphs/prediction_generator.py
lukipuki: Fixed. I've made `fitted_formulas` into a list.

lukipuki (commented):
Protobuf is a typed serialization format, there isn't much to it. Like JSON but much better :)

covid_graphs/covid_graphs/formula.py
```diff
@@ -107,40 +108,82 @@ def formula(x: float):
...
-            xs=xs, ys=country_report.cumulative_active[: until_idx + 1],
-        )
-        # Counterintuitively, `date` + `timedelta` results in `date`.
-        whole_day_offset = np.floor(self.fit.t0)
-        self.fit.shift_forward(whole_day_offset)
+    """
+    The model ('fit') was fitted using data until 'last_data_date'.
+    The trace created by this model starts at 'start_date' shifted by 'fit.t0' days ('fit.t0' is
+    less than 1).
+    """
```
miskosz, line 123: Nit: Move the docstring below line 118.

covid_graphs/covid_graphs/formula.py
```diff
@@ -107,40 +108,82 @@ def formula(x: float):
         )
 
 
+def set_proto_date(proto_date, date: datetime.date) -> None:
```
miskosz: Nit: Usual convention is to prepend functions not exposed outside of the file with `_`. (There are a few more below.)

covid_graphs/covid_graphs/prediction_generator.py
```diff
@@ -7,8 +7,8 @@
 from google.protobuf import text_format  # type: ignore
 
 from .country_report import CountryReport, create_report
-from .formula import FittedFormula
-from .pb.atg_prediction_pb2 import AtgParameters, CountryAtgParameters
+from .formula import fit_country_data
```
miskosz: Nit: I'm a fan of `from . import formula` and then `formula.fit_country_data` since it makes it clear the function comes from a different module.

covid_graphs/covid_graphs/formula.py
```diff
@@ -107,40 +108,82 @@ def formula(x: float):
...
-        whole_day_offset = np.floor(self.fit.t0)
-        self.fit.shift_forward(whole_day_offset)
+    """
+    The model ('fit') was fitted using data until 'last_data_date'.
+    The trace created by this model starts at 'start_date' shifted by 'fit.t0' days ('fit.t0' is
+    less than 1).
+    """
 
-        start_date = date_zero + datetime.timedelta(days=whole_day_offset)
+    def serialize(self):
```
miskosz, line 129: Nit: Maybe `get_atg_parameters`? Also you can set the return type annotation.

Btw I did not originally enable a flag to enforce typing, but we could make it stricter if we wanted.

miskosz (approved):
Looks good! I think this is the right direction. Left only very minor comments.

covid_graphs/covid_graphs/formula.py

covid_graphs/covid_graphs/formula.py
lukipuki: Done. By the way, should this be caught by the linter?

covid_graphs/covid_graphs/formula.py
lukipuki: The type annotations don't work with protos. I think we also tell `mypy` to ignore proto files.

covid_graphs/covid_graphs/prediction_generator.py
lukipuki: Done

covid_graphs/covid_graphs/formula.py
miskosz: It is valid to have a multiline comment in the middle of the class. Maybe we could enforce though having docstrings for all classes and functions; but that might be an overkill for us.

covid_graphs/covid_graphs/formula.py
miskosz: Right. I think that the annotation is though valuable because `mypy` checks type consistency of the functions (even if it does not check the "starting point"), at least I think.

Btw. it is possible to generate protobuf type annotations using some tool (if you'd consider using `mypy` in a prod envirronment).

PR #93, opened 2020-05-08
Filter predictions that differ too much

### LINKED ISSUE #75: Come up with reasonable "prediction statibility" criteria
Predictions for some countries are very unstable, e.g. the best fit created using data until Apr 27 could be very different than the day before or day after. When this happens, it probably means that the prediction won't be very good -- very different curves seem to fit similar data.

We should come up with a precise criterion for curves/traces being too different. For example:

- the days of the peak are too far from each other, or the inflection days from each other.

- the shape of the curve is too different (unclear how to formalize this)

Also, consider random subsets of data for predictions, e.g. dropping a random 1/3rd of data.

lukipuki:
Right now only predictions for the slider are filtered out.

Our only criterion is the distance of the peak, it should not vary too
much.

### DIFF
--- a/covid_graphs/covid_graphs/country_graph.py
+++ b/covid_graphs/covid_graphs/country_graph.py
@@ -2,22 +2,22 @@ import datetime
 import math
 from enum import Enum
 from pathlib import Path
-from typing import List, Tuple
+from typing import Iterable, List, Tuple
 
 import click
 import click_pathlib
 from google.protobuf import text_format  # type: ignore
 from plotly.graph_objs import Figure, Layout, Scatter
 
-from . import formula
+from . import formula, prediction_generator
 from .country_report import CountryReport, create_report
-from .formula import TraceGenerator
+from .formula import FittedFormula, TraceGenerator
 from .pb.atg_prediction_pb2 import CountryAtgParameters
-from .prediction_generator import get_fitted_predictions
 from .predictions import BK_20200329, BK_20200411, CountryPrediction, PredictionEvent, prediction_db
 
 # Extend the predictions at least by 1/5th of the length of active cases.
 EXTENSION_RATIO = 0.2
+MAX_PEAK_DISTANCE = datetime.timedelta(days=14)
 
 
 class GraphType(Enum):
@@ -40,7 +40,7 @@ class GraphAxisType(Enum):
 
 
 def _get_display_range(
-    report: CountryReport, trace_generators: List[TraceGenerator]
+    report: CountryReport, trace_generators: Iterable[TraceGenerator]
 ) -> Tuple[datetime.date, datetime.date]:
     start_date = min(trace_generator.start_date for trace_generator in trace_generators)
 
@@ -54,6 +54,26 @@ def _get_display_range(
     )
 
     return start_date, display_until
+
+
+# TODO: move this to another file
+def _get_earliest_displayable_idx(
+    fitted_formulas: Iterable[FittedFormula], max_peak_distance: datetime.timedelta
+) -> int:
+    peak_days = [
+        formula.get_peak(country_report=None) for formula in reversed(list(fitted_formulas))
+    ]
+
+    min_peak, max_peak = peak_days[0], peak_days[0]
+    take = 1
+    for peak in peak_days[1:]:
+        min_peak = min(min_peak, peak)
+        max_peak = max(max_peak, peak)
+        if max_peak - min_peak <= MAX_PEAK_DISTANCE:
+            take += 1
+        else:
+            return len(peak_days) - take
+    return len(peak_days) - take
 
 
 class CountryGraph:
@@ -83,9 +103,7 @@ class CountryGraph:
             )
             for prediction in country_predictions
         }
-        start_date, display_until = _get_display_range(
-            report, list(trace_generator_by_event.values())
-        )
+        start_date, display_until = _get_display_range(report, trace_generator_by_event.values())
         self.trace_by_event = {
             event: trace_generator.generate_trace(display_until)
             for event, trace_generator in trace_generator_by_event.items()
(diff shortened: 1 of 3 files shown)


lukipuki:
I'm merging this since we need to move on. Let me know if you have comments, they can be fixed later.

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -56,6 +56,26 @@ def _get_display_range(
...
+# TODO: move this to another file
+def _get_earliest_displayable_idx(
+    fitted_formulas: Iterable[FittedFormula], max_peak_distance: datetime.timedelta
+) -> int:
+    peak_days = [
+        formula.get_peak(country_report=None) for formula in reversed(list(fitted_formulas))
+    ]
+
+    min_peak, max_peak = peak_days[0], peak_days[0]
+    take = 1
```
miskosz, line 68: Maybe a more descriptive name can be chosen?

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -56,6 +56,26 @@ def _get_display_range(
     return start_date, display_until
 
 
+# TODO: move this to another file
+def _get_earliest_displayable_idx(
+    fitted_formulas: Iterable[FittedFormula], max_peak_distance: datetime.timedelta
+) -> int:
```
miskosz, line 62: What is the intended logic here? Can you maybe add a comment? (Then as a reviewer I can check that the code does what is intended.)

miskosz: Is it the longest interval ending today where `max_peak - min_peak <= const`? Then maybe `while` is more suitable, sth like:
```py
start_idx = len(peak_days) - 1
max_peak, min_peak = peak_days[-1], peak_days[-1]
while max_peak - min_peak < const and start_idx > 0:
    start_idx -= 1
    max_peak = max(max_peak, peak_days[start_idx])
    min_peak = min(min_peak, peak_days[start_idx])
return start_idx + 1
```

miskosz: Also, I think a more simple condition that the peak is within some distance from the last prediction would be good enough for our purposes.

covid_graphs/covid_graphs/country_graph.py
```diff
@@ -330,25 +348,20 @@ def show_country_plot(country_data_file: Path, prediction_file: Path):
...
-                    prediction_date=formula.last_data_date,
-                ),
-                country=country_report.short_name,
-                formula=formula,
-            )
-            for formula in fitted_formulas
-        ]
+        fitted_predictions = prediction_generator.create_predictions_from_formulas(
+            fitted_formulas[start_idx:], country_report.short_name
+        )
```
miskosz, line 355: There is something suspicious about this (maybe for a different PR). We already have generated predictions (loaded from a file), but we kind of create them again.

covid_graphs/covid_graphs/prediction_generator.py
```diff
@@ -1,35 +1,53 @@
...
 from google.protobuf import text_format  # type: ignore
 
 from . import formula
 from .country_report import CountryReport, create_report
+from .formula import FittedFormula
 from .pb.atg_prediction_pb2 import CountryAtgParameters
 from .predictions import CountryPrediction, PredictionEvent
 
-PREDICTION_DAYS = 21
+PREDICTION_DAYS = 7 * 5
```
miskosz, line 15: Nit: I'd add a comment above saying simply `# Five weeks.`

miskosz (commented):
Looks good. I'm starting to loose track about the stages we add information to formulas (compute start/end dates, add metadata on when the prediction was made, compute the trace), maybe that is something we can simplify going forward.

covid_graphs/covid_graphs/country_graph.py
lukipuki: Added a comment and renamed the function. I have previously thought about your suggested simplification but decided that this one seems better suited for what we want, e.g. in the case of a slider. But I don't have a strong opinion on this one.

I disagree with your code suggestion, it has many issues, e.g. it has many possible off-by-one errors (perhaps not now, but in the future): you return `start_idx+1`, which means you probably chose the wrong invariant. You are also indexing into the array.

I think I might have started with something like what you have but decided on rewriting it.

covid_graphs/covid_graphs/country_graph.py
lukipuki: `take` is a classic functional concept, see e.g.:

- https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.take?view=netcore-3.1

- http://zvon.org/other/haskell/Outputprelude/take_f.html

However, I can rename it to `take_count`.

covid_graphs/covid_graphs/country_graph.py
lukipuki: It's because someone didn't fix the prediction DB yet! :smile:

covid_graphs/covid_graphs/prediction_generator.py
lukipuki: Done

covid_graphs/covid_graphs/country_graph.py
miskosz: I have not encountered it, so feel free to keep it, it was just a suggestion :)

covid_graphs/covid_graphs/country_graph.py
miskosz: The invariant is simple, `start_idx` is the first index not to be included. Maybe a better name can be chosen. I find it tricky to name variables that get updated, since `start_idx` does not have that property until the loop is done. (The same goes also for the original code.) I also like this version since it does not require reversing the array. But your choice.

PR #97, opened 2020-05-09
Bayesian modelling

### LINKED ISSUE #81: Draw credible intervals around the predictions
- Plotly API https://plotly.com/python/v3/continuous-error-bars/

- [PyMC3 might be useful](https://github.com/CamDavidsonPilon/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers) for this endeavour

- The hardest part is probably choosing the right model for the measurement error (loss function) each day.

lukipuki:
We model the number of daily increase of active cases in PyMC3 using a
Laplace distribution (corresponds to L1 loss function). The Bayesian
model converges and the pictures look good, but we have a suspicion that
the Laplace distribution is not the final answer.

### DIFF
--- a/covid_graphs/covid_graphs/bayesian.py
+++ b/covid_graphs/covid_graphs/bayesian.py
@@ -1,7 +1,8 @@
-# Code inspired by Vladimír Boža
+# Code inspired by a program from Vladimír Boža
 
 import random
 from pathlib import Path
+from typing import List
 
 import click
 import click_pathlib
@@ -16,7 +17,7 @@ from .formula import FittedFormula
 from .predictions import CountryPrediction, PredictionEvent
 
 
-def fit_bayesian_model(cases: np.ndarray):
+def fit_bayesian_model(cases: np.ndarray) -> List[AtgModelFit]:
     base_length = 800
     with pm.Model() as model:  # noqa: F841
         days = np.arange(len(cases))
@@ -62,16 +63,20 @@ def fit_bayesian_model(cases: np.ndarray):
 
 @click.command(help="COVID-19 country growth visualization")
 @click.argument("filename", required=True, type=click_pathlib.Path(exists=True))
-def calculate_posterior(filename: Path):
+@click.argument("cutoff", required=False, type=int, default=0)
+def calculate_posterior(filename: Path, cutoff: int):
     report = country_report.create_report(filename)
 
-    cutoff = 1
-    cases = np.diff(report.cumulative_active[:-cutoff])
+    used_length = len(report.dates) - cutoff
+    print(used_length)
+    cases = np.diff(report.cumulative_active[:used_length])
 
     fits = random.sample(fit_bayesian_model(cases), 100)
-    print(fits[:5])
+    print("Five random fits:")
+    for fit in fits[:5]:
+        print(fit)
 
-    last_data_date = report.dates[-cutoff]
+    last_data_date = report.dates[used_length - 1]
     predictions = [
         CountryPrediction(
             prediction_event=PredictionEvent(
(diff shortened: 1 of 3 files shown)


covid_graphs/covid_graphs/bayesian.py
```diff
@@ -0,0 +1,88 @@
+# Code inspired by Vladimír Boža
```
miskosz: 🤣 I think you might want to rephrase.

covid_graphs/covid_graphs/bayesian.py
```diff
@@ -0,0 +1,88 @@
...
+import pymc3 as pm
+
+from . import country_report
+from .country_graph import CountryGraph, GraphType
+from .fit_atg_model import AtgModelFit
+from .formula import FittedFormula
+from .predictions import CountryPrediction, PredictionEvent
+
+
+def fit_bayesian_model(cases: np.ndarray):
```
miskosz: Nit: Return type.

covid_graphs/covid_graphs/bayesian.py
```diff
@@ -0,0 +1,88 @@
...
+    base_length = 800
+    with pm.Model() as model:  # noqa: F841
```
miskosz, line 22: `with pm.Model():`

miskosz: Actually the whole construct seems a bit fishy, do you know why do we need the `with` line (what does it do)? Anyway, this is experimental functionality so we don't have to solve it.

covid_graphs/covid_graphs/bayesian.py
```diff
@@ -0,0 +1,88 @@
...
+    ]
+
+
+@click.command(help="COVID-19 country growth visualization")
+@click.argument("filename", required=True, type=click_pathlib.Path(exists=True))
+def calculate_posterior(filename: Path):
+    report = country_report.create_report(filename)
+
+    cutoff = 1
+    cases = np.diff(report.cumulative_active[:-cutoff])
```
miskosz: Cutoff -1 is a strange value, you use all the datapoints except the last one.

miskosz (approved):
LGTM

covid_graphs/covid_graphs/bayesian.py
lukipuki: Done

covid_graphs/covid_graphs/bayesian.py
lukipuki: Done

covid_graphs/covid_graphs/bayesian.py
lukipuki: `with pm.Model():` doesn't work. The authors explain it here:
https://nbviewer.jupyter.org/github/CamDavidsonPilon/Probabilistic-Programming-and-Bayesian-Methods-for-Hackers/blob/master/Chapter2_MorePyMC/Ch2_MorePyMC_PyMC3.ipynb#Model-Context

I don't think we have a lot of freedom here.

covid_graphs/covid_graphs/bayesian.py
lukipuki: It should be an optional command line parameter, so I've made it that. And I've changed the default value to 0.

miskosz: It should be `len(report.cumulative_active)` instead of `0` in my opinion.

lukipuki: This is how it's used, so value of 0 is fine. But perhaps the name `cutoff` is not the best, since you misunderstood.
```
    used_length = len(report.dates) - cutoff
    cases = np.diff(report.cumulative_active[:used_length])
```

miskosz: Oh, you updated the code... No it is completely fine, just I though it stayed `cases = np.diff(report.cumulative_active[:-cutoff])`.

PR #98, opened 2020-05-09
Python package management. Test covid_web in CI.

### LINKED ISSUE #90: Single python venv
Covid_graphs and covid_web packages has it's own python virtual envs, but web has graphs as dependency. We should consider making one shared virtualenv.

miskosz:
Created a `python` directory that contains two packages:
* `covid_graphs` (moved from `/covid_graphs`)
* `covid_web` (moved from `/web/covid_web`)

Testing scripts were moved from `/covid_graphs` to `/python` and made to work with both packages.


lukipuki (approved):

miskosz:
It is just a draft for now, I need to update Readme and test that docker is not broken :)

rejdi:
@miskosz english please ;)
Also do we plan to keep react web under python-web? I'd suggest adding it to it's own directory `/react-web`

miskosz:
I kept `react-web` with `nginx` it the `web` directory as it was. Do you suggest to pull it out?

rejdi:
No, I overlooked it. I thought that it was moved together with python-web.

miskosz:
The `Readme`s are a bit inconsistent in the end (I put docker usage to the python `Readme`), but we can fix that later.

Issue #95, opened 2020-05-08
Replace Python REST endpoints by statically generated JSON files served by Nginx


rejdi:

lukipuki:
Blocked by https://github.com/lukipuki/COVID-19-simulation/issues/94

PR #101, opened 2020-05-09
REST refactoring. Added links to list for all available predictions.

rejdi:


python/covid_web/covid_web/rest.py
```diff
@@ -79,7 +82,34 @@ def get_app(self):
         return self
 
     def get_available_predictions(self):
-        return jsonify(self.available_predictions)
+        result = []
+        for prediction in self.available_predictions:
+            prediction['link'] = url_for("covid19_get_specific_prediction", country=prediction['country'], prediction=prediction['prediction'])
```
lukipuki: The CI will tell you to format this line :)

python/covid_web/covid_web/rest.py
```diff
@@ -79,7 +82,34 @@ def get_app(self):
...
+        if len(result) == 0:
+            abort(404)
+
+        return jsonify(result)
+
+    def get_predictions_by_name(self, date: str):
+        result = []
+        for prediction in self.predictions:
+            if prediction["date_name"] == date:
+                result.append(prediction)
```
lukipuki: We prefer list comprehensions for these
```suggestion
    result = [
        prediction
        for prediction in self.predictions:
        if prediction["date_name"] == date
    ]
```

python/covid_web/covid_web/rest.py
```diff
@@ -1,7 +1,7 @@
 from pathlib import Path
 from typing import Dict, List, Tuple
 
-from flask import Flask, abort, jsonify
+from flask import Flask, abort, jsonify, url_for
```
lukipuki:
```suggestion
from flask import Flask
import flask
```
And then use `flask.abort`, `flask.jsonify`, etc.

python/covid_web/covid_web/rest.py
```diff
@@ -79,7 +82,34 @@ def get_app(self):
         return self
 
     def get_available_predictions(self):
-        return jsonify(self.available_predictions)
+        result = []
+        for prediction in self.available_predictions:
```
lukipuki: Since available predictions are created in the constructor, why isn't the link set then?

python/covid_web/covid_web/rest.py
```diff
@@ -79,7 +82,34 @@ def get_app(self):
...
+            prediction['link'] = url_for("covid19_get_specific_prediction", country=prediction['country'], prediction=prediction['prediction'])
+            result.append(prediction)
+
+        return jsonify(result)
+
+    def get_predictions_by_country(self, country: str):
+        result = []
+        for prediction in self.predictions:
+            if prediction["short_name"] == country:
```
lukipuki: Add a list comprehension here as well.

python/covid_web/covid_web/rest.py
rejdi: Links are not working at that stage. I got
`RuntimeError: Attempted to generate a URL without the application context being pushed. This has to be executed when application context is available.`

lukipuki: Then I think we should do it differently, because this is inefficient. You create links every time you call this function.

- Make `_create_available_predictions` public, so it becomes `Rest.create_available_predictions()`. It will include link creation.
- Call it in `server.py` right before `def covid19_available_predictions()`.
- `def covid19_get_specific_prediction(prediction, country)` should be moved up in `server.py`, before `covid19_available_predictions`.

lukipuki: Or there's another way: do it only once. Check if the available predictions are initialized and if not, initialize them.

rejdi: I seems that it's still too soon. It looks like `url_for` is available when Flask variable `SERVER_NAME` is set. This is done during execution, but it also can be done before it, by manually setting it. It seems to be quite hacky. I suggest just removing it (we can live without it) or creating it lazily once during the first call.

ref: https://stackoverflow.com/questions/31766082/flask-url-for-error-attempted-to-generate-a-url-without-the-application-conte

lukipuki: I should perhaps also ask why this isn't instead stored offline and served via Nginx, as we plan to do at some point: https://github.com/lukipuki/COVID-19-simulation/issues/95

python/covid_web/covid_web/rest.py
rejdi: Changed.

python/covid_web/covid_web/rest.py
rejdi: Changed.

python/covid_web/covid_web/rest.py
rejdi: Changed.

python/covid_web/covid_web/rest.py
rejdi: Applied.

python/covid_web/covid_web/rest.py
rejdi: Another PR.

python/covid_web/covid_web/rest.py
```diff
@@ -79,24 +80,51 @@ def get_app(self):
         return self
 
     def get_available_predictions(self):
-        return jsonify(self.available_predictions)
+        # lazy link initialization
+        if 'link' not in self.available_predictions[0]:
```
lukipuki: Now it occurred to me that you might have a concurrency problem here ... Not very likely though.

lukipuki (approved):

python/covid_web/covid_web/rest.py
rejdi: I removed this feature for now.

rejdi:
All suggestions are already implemented in #103. Closing this one.

PR #102, opened 2020-05-09
Load daily predictions into PredictionDb.

### LINKED ISSUE #94: Add daily predictions to prediction_db
So the web infrastructure may easily use daily predictions.

miskosz:
- Add `predictions.load_prediction_db(prediction_dir)` function to instantiate `PredictionDb` object
- Add `prediction_dir` parameter to cli commands
- Add `py.typed` file to enable type-checking across packages (https://mypy.readthedocs.io/en/stable/installed_packages.html)
- Bugfix: Enable `mypy` for `covid_web`

Next steps:
It seems to me we should add a flag to `PredictionEvent` to distinguish manual and automatic predictions. Also I think that the whole data stored in PredictionDb should be stored in proto files. Probably :)

### DIFF
--- a/python/covid_graphs/covid_graphs/country_graph.py
+++ b/python/covid_graphs/covid_graphs/country_graph.py
@@ -315,7 +315,7 @@ class CountryGraph:
     "country_data_file", required=True, type=click_pathlib.Path(exists=True),
 )
 @click.argument(
-    "prediction_dir", required=False, type=click_pathlib.Path(exists=True),
+    "prediction_dir", required=True, type=click_pathlib.Path(exists=True),
 )
 def show_country_plot(country_data_file: Path, prediction_dir: Path):
     country_report = create_report(country_data_file)
--- a/python/covid_graphs/covid_graphs/country_graph.py
+++ b/python/covid_graphs/covid_graphs/country_graph.py
@@ -6,14 +6,12 @@ from typing import Iterable, List, Tuple
 
 import click
 import click_pathlib
-from google.protobuf import text_format  # type: ignore
 from plotly.graph_objs import Figure, Layout, Scatter
 
-from . import formula, prediction_generator
+from . import predictions
 from .country_report import CountryReport, create_report
-from .formula import FittedFormula, TraceGenerator
-from .pb.atg_prediction_pb2 import CountryAtgParameters
-from .predictions import BK_20200329, BK_20200411, CountryPrediction, PredictionEvent, prediction_db
+from .formula import TraceGenerator
+from .predictions import BK_20200329, BK_20200411, CountryPrediction, PredictionEvent
 
 # Extend the predictions at least by 1/5th of the length of active cases.
 EXTENSION_RATIO = 0.2
@@ -54,30 +52,6 @@ def _get_display_range(
     )
 
     return start_date, display_until
-
-
-# TODO: move this to another file
-def _calculate_earliest_displayable_idx(
-    fitted_formulas: Iterable[FittedFormula], max_peak_distance: datetime.timedelta
-) -> int:
-    """
-    Calculates the earliest index in fitted_formulas, after which the peaks of formulas are within
-    'max_peak_distance'.
-    """
-    peak_days = [
-        formula.get_peak(country_report=None) for formula in reversed(list(fitted_formulas))
-    ]
-
-    min_peak, max_peak = peak_days[0], peak_days[0]
-    take_count = 1
-    for peak in peak_days[1:]:
-        min_peak = min(min_peak, peak)
-        max_peak = max(max_peak, peak)
-        if max_peak - min_peak <= max_peak_distance:
-            take_count += 1
-        else:
-            break
-    return len(peak_days) - take_count
 
 
 class CountryGraph:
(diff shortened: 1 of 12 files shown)


python/covid_web/covid_web/rest.py
```diff
@@ -3,24 +3,26 @@
```
lukipuki, line 3: There will be a lot of conflicts in this file with https://github.com/lukipuki/COVID-19-simulation/pull/101, we should resolve the conflicts carefullly.

python/covid_graphs/covid_graphs/country_graph.py
```diff
@@ -6,14 +6,12 @@
...
+from . import predictions
 from .country_report import CountryReport, create_report
-from .formula import FittedFormula, TraceGenerator
-from .pb.atg_prediction_pb2 import CountryAtgParameters
-from .predictions import BK_20200329, BK_20200411, CountryPrediction, PredictionEvent, prediction_db
+from .formula import TraceGenerator
+from .predictions import BK_20200329, BK_20200411, CountryPrediction, PredictionEvent
 
 # Extend the predictions at least by 1/5th of the length of active cases.
 EXTENSION_RATIO = 0.2
```
lukipuki, line 17: The `MAX_PEAK_DISTANCE` constant is now unused (I know we would eventually want to move it back)

python/covid_graphs/covid_graphs/predictions.py
```diff
@@ -196,5 +204,77 @@ def predictions_for_event(self, prediction_event: PredictionEvent) -> List[Count
     def predictions_for_country(self, country: str) -> List[CountryPrediction]:
         return [p for p in self._prediction_database if p.country == country]
 
+    def select_predictions(
+        self, country: str, last_data_dates: List[datetime.date]
+    ) -> List[CountryPrediction]:
+        return [
+            p
+            for p in self._prediction_database
+            if p.country == country and p.prediction_event.last_data_date in last_data_dates
```
lukipuki, line 213: I think we should have a dictionary of predictions per country to speed this up.

Also, isn't `in last_data_dates` too slow? Or we don't care? I'd at least convert it to a `set`.

python/covid_graphs/covid_graphs/predictions.py
```diff
@@ -196,5 +204,77 @@ def predictions_for_event(self, prediction_event: PredictionEvent) -> List[Count
...
+        start_idx = _calculate_earliest_displayable_idx(fitted_formulas, MAX_PEAK_DISTANCE)
+        fitted_predictions = _create_predictions_from_formulas(
+            fitted_formulas[start_idx:], country_atg_parameters.short_country_name
+        )
+        country_predictions.extend(fitted_predictions)
+
+    return PredictionDb(country_predictions=country_predictions)
+
+
+# TODO(miskosz): Move to the caller.
```
lukipuki: Can you clarify what you mean by moving to the caller?

python/covid_web/covid_web/country_dashboard.py
```diff
@@ -110,7 +111,7 @@ def create_dashboard(self, dashboard_type: DashboardType, server: Flask) -> dash
 
     @staticmethod
     def _create_graphs(
-        report_by_short_name: Dict[str, CountryReport], prediction_event: PredictionEvent,
+        prediction_db: PredictionDb, report_by_short_name: Dict[str, CountryReport], prediction_event: PredictionEvent,
```
lukipuki, line 114: Alternatively, this wouldn't be a static method anymore and would use `self.prediction_db`.

python/covid_web/covid_web/country_dashboard.py
```diff
@@ -170,16 +171,16 @@ def _create_buttons(
...
-        # TODO: Predictions will be generated statically in the future.
         graphs_by_country = {
             country_short_name: CountryGraph(
                 self.report_by_short_name[country_short_name],
-                get_fitted_predictions(
-                    self.report_by_short_name[country_short_name],
+                # TODO(mszabados): Make it possible to select only automatic predictions.
+                self.prediction_db.select_predictions(
+                    country=country_short_name,
                     last_data_dates=self.report_by_short_name[country_short_name].dates[-1:],
```
lukipuki: The range here was really short, so that the prediction computation doesn't take ages. We can extend it after your change.
```suggestion
                    last_data_dates=self.report_by_short_name[country_short_name].dates[-28:],
```

python/covid_web/covid_web/country_dashboard.py
```diff
@@ -193,15 +194,12 @@ def update_graph(graph_axis_type_str, country_short_name):
...
+        for country_short_name in self.prediction_db.get_countries():
             report = self.report_by_short_name[country_short_name]
-            country_predictions = prediction_db.predictions_for_country(country=country_short_name)
-            country_predictions.extend(
-                get_fitted_predictions(
-                    country_report=report, last_data_dates=[report.dates[-1], report.dates[-8]]
-                )
+            country_predictions = self.prediction_db.select_predictions(
+                country=country_short_name,
+                last_data_dates=[report.dates[-1], report.dates[-8]],
```
lukipuki: So the `/daily` endpoint will only show two predictions, no Boďová & Kollár?

If that's the case I'd prefer also showing either `-15`. Or every 5-th day: `-11, -6, -1`.

lukipuki (approved):
Nicely done! Much needed and helpful functionality.

python/covid_web/covid_web/rest.py
rejdi: I'm closing #101 for now and I'll handle them once this gets merged. Meanwhile, you may review updated version https://github.com/lukipuki/COVID-19-simulation/pull/103

lukipuki:
There should be an option to only return the automated predictions, we don't want BK predictions in the slider: Screenshot_20200510_085702

python/covid_graphs/covid_graphs/country_graph.py
miskosz: I'm not sure what is the suggestion here, the constant has moved :)

python/covid_graphs/covid_graphs/predictions.py
miskosz: I don't think this is the bottleneck (graph generation is). We can optimise in another PR if we want to.

python/covid_graphs/covid_graphs/predictions.py
miskosz: I'll rephrase. I don't think that PredictionDb should filter data that is available, it should serve everything and the user should choose what they want. Choosing which traces to display can be responsibility of the dashboard; or we can move it to prediction generation (so that uncertain predictions are not even exported). My preference would be the former.

python/covid_web/covid_web/country_dashboard.py
miskosz: You know I like it pure :)

python/covid_web/covid_web/country_dashboard.py
miskosz: I think we should get rid of the endpoint now that we have the slider, what do you think?

python/covid_graphs/covid_graphs/country_graph.py
lukipuki: It's not moved, it's still on line 20

python/covid_graphs/covid_graphs/predictions.py
lukipuki: The dictionary is a bigger change, but making it into a set is more important and just a one/two line change.

It could be a bottleneck with 28 days selected (see my other suggestion below).

python/covid_graphs/covid_graphs/predictions.py
lukipuki: Right, I agree, but please expand the TODO a bit.

python/covid_web/covid_web/country_dashboard.py
lukipuki: I actually like it, now that we've removed the BK predictions from it! It serves a different purpose than the slider, you can see more predictions at once.

But I agree that it's added complexity in the code which needs to be maintained.

python/covid_web/covid_web/country_dashboard.py
miskosz: We can have all with the slider.

python/covid_graphs/covid_graphs/country_graph.py
miskosz: I have no idea how is this possible :D Not only I thought I moved it, but also I did not notice that it stayed here on github.

python/covid_web/covid_web/country_dashboard.py
miskosz: I brought the BK predictions back so that the functionality is not changed... 😇 I think I'll just commit, feel free to change the functionality again.

python/covid_graphs/covid_graphs/predictions.py
miskosz: I measured it; there's no noticeable effect.

PR #103, opened 2020-05-10
REST interface refactored with command line static generator

### LINKED ISSUE #95: Replace Python REST endpoints by statically generated JSON files served by Nginx

rejdi:
This PR refactor REST interface and allow consumers to fetch multiple predictions at once either by country or by prediction name.
This PR also tries to implement and closes #95, so that we should be able to run without python REST backend and save some memory and CPU.
This PR is blocked by https://github.com/lukipuki/COVID-19-simulation/issues/94


docker-compose.yml
```diff
@@ -24,6 +25,14 @@ services:
     volumes:
       - "react-web:/react-web"
     command: "bash -c ' rm -rf /react-web/* && cp -a . /react-web/'"
+  generate-static-rest:
+    build: .
+    image: flask_server
+    volumes:
+      - "./data:/covid19/data:ro"
+      - "static-rest:/covid19/rest"
+    command: "covid_web.generate_static_rest /covid19/data/ /covid19/rest/"
```
lukipuki: I think another container is an overkill. I'd instead create an entry point for the `flask_server` container with 
```
#!/bin/sh
covid_web.generate_static_rest /covid19/data/ /covid19/rest/
uwsgi --uid www-data --gid www-data --socket 0.0.0.0:5000 --die-on-term -w covid_web.wsgi:app
```

python/covid_web/covid_web/rest.py
```diff
@@ -79,24 +98,103 @@ def get_app(self):
...
+        return flask.jsonify(result)
+
+    def _get_predictions_by_name(self, prediction_name: str):
+        result = [
+            prediction
+            for prediction in self.predictions
+            if prediction["prediction_name"] == prediction_name
+        ]
+
+        return result
```
lukipuki:
```suggestion
        return [
            prediction
            for prediction in self.predictions
            if prediction["prediction_name"] == prediction_name
        ]
```

python/covid_web/covid_web/rest.py
```diff
@@ -79,24 +98,103 @@ def get_app(self):
...
-            abort(404)
+            flask.abort(404)
 
-        return jsonify(self.country_reports_active[country])
+        return flask.jsonify(self.country_reports_active[country])
 
+    def generate_static_files(self, output_dir: Path) -> None:
+        Path(output_dir / f"data").mkdir(parents=True, exist_ok=True)
+        Path(output_dir / f"predictions/by_prediction").mkdir(parents=True, exist_ok=True)
+        Path(output_dir / f"predictions/by_country").mkdir(parents=True, exist_ok=True)
```
lukipuki, line 146:
```suggestion
        Path(output_dir / "data").mkdir(parents=True, exist_ok=True)
        Path(output_dir / "predictions" / "by_prediction").mkdir(parents=True, exist_ok=True)
        Path(output_dir / "predictions" / "by_country").mkdir(parents=True, exist_ok=True)
```

... and similarly at other places in this function.

lukipuki (approved):

docker-compose.yml
rejdi: What about generating those files in `download_all.zsh`?

python/covid_web/covid_web/rest.py
rejdi: Applied.

python/covid_web/covid_web/rest.py
rejdi: Applied.

docker-compose.yml
lukipuki: That's fine too

lukipuki: I don't think that's a good place to do it, that script is just for downloading data. What you have here is data transformed into something else. For example you'd have to run `download_all.zsh` every time you change something in your code, which doesn't make sense conceptually.

The entry point is the right place to do it IMHO.

rejdi: `download_all.zsh` is not just downloading the data. It's also generating data and prediction protobuf files. If you modify something in `generate_predictions` you have to rerun script or post-processing command. It also make sense to put it here, as static REST files needs to be regenerated only when we have new data or updated predictions. If I'm not missing something, that's done only in `download_all.zsh`.

I suggest to split `download_all.zsh` into two files. One for downloading and second for post-processing (predictions + rest). `download_all.zsh` would also call postprocessing script.

lukipuki: From the point of view of the graphs, the predictions are just data. In a hypothetical app with a database, you'd put country data into a Postgres DB together with the predictions. The predictions would be calculated every night and written into the database.

However, you wouldn't put JSON files into the database (at least I don't see a reason for doing that).

rejdi: Yep, JSON in database does not make sense. Applied new entry point solution, since it's running fast enough, it can be executed every time we start the flask server.

python/run_tests.sh
```diff
@@ -7,20 +7,23 @@ then
...
-
-echo "👮 Type checking..."
-mypy $PYTHON_PACKAGES
-
-echo "🧶 Linting..."
-flake8 --config setup.cfg --jobs auto $PYTHON_PACKAGES
-
-echo "🏃 Running tests..."
-pytest
+for PYTHON_PACKAGE in "${PYTHON_PACKAGES[@]}"
```
lukipuki: What's the reason for the for loop?

python/covid_web/covid_web/server.py
```diff
@@ -55,13 +55,22 @@ def _create_rest(data_dir: Path, server: Flask):
...
+        return rest.get_predictions_by_name(prediction)
+
+    @server.route("/covid19/rest/predictions/by_country/<country>")
+    def covid19_get_predictions_by_country(country):
+        return rest.get_predictions_by_country(country)
+
+    @server.route("/covid19/rest/predictions/by_prediction/<prediction>/<country>")
+    @server.route("/covid19/rest/predictions/by_country/<country>/<prediction>")
+    def covid19_get_specific_prediction(prediction, country):
+        return rest.get_specific_prediction(prediction, country)
```
lukipuki: Wait, isn't this PR about moving the REST endpoints out of Flask?

python/covid_web/covid_web/server.py
```diff
@@ -63,13 +62,22 @@ def _create_rest(server: Flask, data_dir: Path, prediction_dir: Path):
...
 
-    @server.route("/covid19/rest/predictions/list/")
+    @server.route("/covid19/rest/predictions/list")
     def covid19_available_predictions():
         return rest.get_available_predictions()
 
-    @server.route("/covid19/rest/predictions/data/<date>/<country>")
-    def covid19_get_specific_prediction(date, country):
-        return rest.get_specific_prediction(date, country)
+    @server.route("/covid19/rest/predictions/by_prediction/<prediction>")
```
lukipuki: Aren't we getting rid of the Flask endpoints?

lukipuki (approved):

Dockerfile
```diff
@@ -9,3 +9,4 @@ RUN apt update \
 COPY python /python
 RUN pip3 install /python/covid_graphs /python/covid_web && rm -rf /python
 RUN pip3 install uwsgi
+COPY python/covid_web/flask_server.sh /usr/bin/
```
lukipuki, line 12: Should we mark it as `ENTRYPOINT`?

python/run_tests.sh
rejdi: Old version was checking just `covid_graphs`.

python/covid_web/covid_web/server.py
rejdi: I didn't plan to do this step in this PR, however I can do it now.

python/covid_web/covid_web/server.py
rejdi: Same as previous.

Dockerfile
rejdi: Marked.

python/covid_web/covid_web/server.py
rejdi: Removed from flask.

python/format_code.sh
```diff
@@ -8,8 +8,12 @@ fi
 
 PYTHON_PACKAGES=(covid_graphs covid_web)
 
-echo "Formatting imports (isort)"
-isort --apply --recursive $PYTHON_PACKAGES
-
-echo "Formatting code (black)"
-black $PYTHON_PACKAGES
+for PYTHON_PACKAGE in "${PYTHON_PACKAGES[@]}"
```
lukipuki: Paging @miskosz , so that he takes a look.

rejdi: Created separate issue #110

PR #104, opened 2020-05-10
Polish UI

### LINKED ISSUE #92: Polish the new UI
- [x] The new light green zone has a tooltip and there's another redundant one:

Screenshot_20200508_120749

- [ ] The legend is too big:

image_from_ios

- [x] We currently extend the graph by 7 days (`EXTENSION_PERIOD` in the code) after active cases end, but this should be longer at least on some graphs. Either increase this constant or scale it with the length of `active_case`. E.g. make it at least 20% more than the length of `active_cases`.

lukipuki:
* Slider is on top of the graph for easier discoverability
* Blue color is back for our automated predictions
* I've updated the Markdown text
* Slovakia doesn't shift the data by 1 day anymore

Slider UI now:
Screenshot_20200510_091620

This needs to wait until #102 is merged.

### DIFF
--- a/python/covid_graphs/covid_graphs/country_graph.py
+++ b/python/covid_graphs/covid_graphs/country_graph.py
@@ -10,7 +10,7 @@ from plotly.graph_objs import Figure, Layout, Scatter
 
 from . import predictions
 from .country_report import CountryReport, create_report
-from .formula import TraceGenerator
+from .formula import FittedFormula, TraceGenerator
 from .predictions import BK_20200329, BK_20200411, CountryPrediction, PredictionEvent
 
 # Extend the predictions at least by 1/5th of the length of active cases.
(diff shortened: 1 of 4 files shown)


python/covid_graphs/covid_graphs/country_graph.py
```diff
@@ -126,8 +126,10 @@ def _create_visibility_vector(event_name: str) -> List[bool]:
         return dict(
             active=len(steps) - 1,
             currentvalue={"prefix": "Prediction date: "},
-            pad={"t": 60},
+            pad={"b": 10},
```
rejdi: If the screenshot you've attached is up2date I'd add greater bottom padding.

rejdi (approved):

miskosz:
Btw. let's get rid of the "Date" label at the bottom. I know that I wanted graph axis labeled, but actually it is super obvious and the label just gets in the way.

python/covid_graphs/covid_graphs/country_graph.py
lukipuki: Bumped it to 20.

lukipuki:
Got rid of the date label, let's hope people won't make this joke: https://xkcd.com/833/ (I've made it about 3 times in the last month regarding COVID-19 graphs).

PR #109, opened 2020-05-10
React web with new features

rejdi:
This is larger refactor of react-web. Following features are included:
* Adopted new REST interface from #103 
* Added slider + play/pause
* Bugfixes...
PR #103 already included.

react-web


lukipuki:
How can I try this out locally?

rejdi:
* `docker-compose up`, http://localhost:8081/covid19/graphs/
or
* in web/react-web/ run `npm install`, `npm run start` (development mode) + run backend (with reverted f84dd5c). http://localhost:3000/

rejdi:
Rebased on master.

lukipuki:
> * `docker-compose up`, http://localhost:8081/covid19/graphs/

This doesn't render anything in my Chrome. Same for graphs.lukipuki.sk/covid19/graphs.

lukipuki:
My comments from highest priority to lowest, but most of these are huge endeavors, so should be next PRs.

- "Results" should be the default tab, you read "About" maybe once and then you are done.
- Plotly.js for graphs, as Plotly does it better (though HighCharts seems to be more responsive)
- Per capita scaling of graphs
- I think we should start with showing something instead of an empty page. We can pick Slovakia or some other country. The Python web defaults on Italy.
- I'm not a fan of pastel colors on the web [actually anywhere :grin: ], I'd rather have the left part white or some other color.

web/react-web/src/Sidebar/index.js
```diff
@@ -23,88 +21,75 @@ class Sidebar extends Component {
...
+                        <>
+                            <div className='buttons'>
+                                <button className={activeTab === 0 ? 'active' : ''} onClick={this.selectTab(0)}>About</button>
+                                <button className={activeTab === 1 ? 'active' : ''} onClick={this.selectTab(1)}>Results</button>
+                            </div>
+                            {activeTab === 0 &&
+                                //TODO: load markdown content from backend and render it, once it'll be available
+                                <>
+                                    <h1>COVID-19 predictions of Boďová and Kollár</h1>
+                                    <p>Mathematicians Katarína Boďová and Richard Kollár predicted in March and April 2020
```
lukipuki, line 39: By the way, we have clarified this text recently in the Python code base. We should share the text between the projects, I think Python can provide the HTML via a REST call. Can be done in another PR.

rejdi: Either markdown or html. No problem.

lukipuki (approved):

PR #111, opened 2020-05-11
Static REST readme

rejdi:


python/README.md
```diff
@@ -32,4 +32,9 @@ covid_graphs.show_scatter_plot ../simulation/Slovakia.data polynomial.sim
 covid_graphs.show_heat_map exponential.sim
 ```
 
+To create static data used for our REST service:
```
lukipuki, line 35: Shouldn't this be placed in the `/web/README.md` file? (Not sure if it exists)

rejdi: It's mentioned in another PR
https://github.com/lukipuki/COVID-19-simulation/blob/new-web-slider/web/react-web/README.md

lukipuki: Then I think we don't need to mention it here. There's little utility of the generated JSON's without the React web.

E.g. above we mention graphing scripts that have directly visible outputs, useful on their own. The output of `generate_static_rest` is only useful once you start the web app.

rejdi: I disagree, since it's a part of package and it deserve some comment. Moreover, if you are not python user and you are trying to locally run react-web, you are pointed here, where you also get instructions on how to work with python package (create venv, activate, pip install...)

lukipuki: Okay, I guess it's better to have more documentation.

PR #116, opened 2020-05-12
Add population field to CountryData

### LINKED ISSUE #114: Gather population for each country
Find the reliable population numbers for each country and add it to source data, so we can scale predictions per capita.

lukipuki:


simulation/src/country_data.proto
```diff
@@ -17,5 +17,6 @@ message DailyStats {
 message CountryData {
   string name = 1;
   string short_name = 3;
+  uint32 population = 4;
```
rejdi: I suggest uint64 if we would like to add whole world.

lukipuki: Done

rejdi (approved):

PR #118, opened 2020-05-13
Clip peaks that are too far

lukipuki:


python/covid_graphs/covid_graphs/predictions.py
```diff
@@ -241,36 +243,50 @@ def load_prediction_db(prediction_dir: Path) -> PredictionDb:
...
-    Calculates the earliest index in fitted_formulas, after which the peaks of formulas are within
-    'max_peak_distance'.
+    Returns a half-open interval [start, end), for which the following holds:
+    * fitted_formulas[end - 1] is the latest prediction whose peak is not too far in the future,
+      i.e. it's earlier than 'max_peak_distance' from now.
+    * fitted_formula[start] is the earliest prediction, after which the peaks are within
+      'max_peak_variability'.
     """
     peak_days = [
         formula.get_peak(country_report=None) for formula in reversed(list(fitted_formulas))
```
miskosz: I would prefer if we did not reverse the array, it makes the code more difficult to follow (you can iterate in reverse order by using `array[from:to:step]` notation with a negative step)

python/covid_graphs/covid_graphs/predictions.py
```diff
@@ -241,36 +243,50 @@ def load_prediction_db(prediction_dir: Path) -> PredictionDb:
...
     ]
+    skip = 0
+    latest_peak = datetime.datetime.now() + max_peak_distance
+    while skip < len(peak_days) and peak_days[skip] > latest_peak:
+        skip += 1
+    clipped_peaks = peak_days[skip:]
+    if len(clipped_peaks) == 0:
+        return skip, skip
```
miskosz: Nit: Consider `if skip == len(peak_days):` on the line before clipping.

miskosz (approved):
I have trouble wrapping my head around reversing, clipping and index counting, but I trust you that it works. (You can consider writing a test to earn bonus points.)

python/covid_graphs/covid_graphs/predictions.py
lukipuki: I'm thinking that the API might not be the best right now, it was made for the previous use case which lived outside of `prediction.py`.

I suggest
- accepting `FittedFormula`'s in any order and we will sort them internally however we want.
- instead of an interval, we will return a subset of `fitted_formulas` with the desired properties.

This will simplify things and we would just construct the list instead of calculating the interval.

rejdi: That will definitely help!

python/covid_graphs/covid_graphs/predictions.py
```diff
@@ -241,36 +243,50 @@ def load_prediction_db(prediction_dir: Path) -> PredictionDb:
...
+    for peak in clipped_peaks[1:]:
         min_peak = min(min_peak, peak)
         max_peak = max(max_peak, peak)
-        if max_peak - min_peak <= max_peak_distance:
+        if max_peak - min_peak <= max_peak_variability:
             take_count += 1
         else:
             break
-    return len(peak_days) - take_count
+    return len(clipped_peaks) - take_count, len(clipped_peaks)
```
rejdi: Are clipped peaks monotone? If not, you may skip some predictions that are within interval!

python/covid_graphs/covid_graphs/predictions.py
lukipuki: I've now changed the API, PTAL.

I'm not inclined to simplify this further, since:
* this is really a TopCoder division 2 500 point problem, maybe not even that
* there are other parts of the code that need love right now :smile:

python/covid_graphs/covid_graphs/predictions.py
lukipuki: I've rewritten the code, so I'm not sure if this comment still applies.

python/covid_graphs/covid_graphs/predictions.py
lukipuki: How do you mean your "you may"?

- "it's a bug, please don't do it" - I think this interpretation isn't likely, since we are returning a full interval without skipping predictions within the interval.
- "please create holes in the interval" - I think this would look weird on the slider.

rejdi: Ok, now I understand what was the point.

lukipuki:
Forgot to say that I've also added a test.

rejdi (approved):

PR #119, opened 2020-05-14
Extend rest+about page

rejdi:
Added population to static REST.
Export and use shared text on About page.

### DIFF
--- a/python/covid_web/covid_web/rest.py
+++ b/python/covid_web/covid_web/rest.py
@@ -1,3 +1,4 @@
+import shutil
 from pathlib import Path
 from typing import Dict, List, Tuple
 
@@ -9,7 +10,8 @@ from covid_graphs import predictions
 from covid_graphs.country_graph import CountryGraph
 from covid_graphs.country_report import CountryReport, create_report
 from covid_graphs.predictions import PredictionDb
-from covid_web import country_dashboard
+
+CURRENT_DIR = Path(__file__).parent
 
 
 @click.command(help="COVID-19 static REST generator")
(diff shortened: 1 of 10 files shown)


python/covid_web/covid_web/rest.py
```diff
@@ -145,6 +148,13 @@ def generate_static_files(self, output_dir: Path) -> None:
         Path(output_dir / "predictions/by_prediction").mkdir(parents=True, exist_ok=True)
         Path(output_dir / "predictions/by_country").mkdir(parents=True, exist_ok=True)
 
+        about = "\n".join(
+            line.strip() for line in country_dashboard.get_about_content().splitlines()
```
lukipuki: I have mixed feelings about exposing `country_dashboard` like this, but I'm not used to creating UI at work. The about content sounds like something that should live outside of `country_dashboard.py`.

Also, why are you doing all the `.strip()` and `.splitlines()`?

lukipuki (approved):

python/covid_web/covid_web/rest.py
rejdi: Actually about content should live in it's own file as a resource, instead of hardcoded in the source. I'm not sure, where to place it. I would to put it beside index.html and then load in `country_dashboard` and static rest generator.

Those split&strip is here, because python's multiline string has heading space on each line. Markdown library I'm using, render such markdown differently than the one used by covid_web. Removing those space helped.

lukipuki: Yeah, having it besides `index.html` in an external file sounds good.

I think moving it to the external file will also solve the `strip + splitlines` issue.

Issue #113, opened 2020-05-11
Improvements for react-web


rejdi:
- [x] "Results" should be the default tab, you read "About" maybe once and then you are done.
- [x] The url `graphs.lukipuki.sk/covid19/graphs` is redundant and different from the rest. I suggest `/covid19/predictions/overview` or similar. The name could be whatever, but should be under `/predictions`.
- [x] "About" page should use shared text with python
- [x] Plotly.js for graphs, as Plotly does it better (though HighCharts seems to be more responsive)
- [x] Per capita, same peak, none - scaling of graphs
- [x] I think we should start with showing something instead of an empty page. We can pick Slovakia or some other country. The Python web defaults on Italy.
- [x] Make the URL work without the trailing slash.
- [x] I'm not a fan of pastel colors on the web [actually anywhere :grin: ], I'd rather have the left part white or some other color.

PR #122, opened 2020-05-14
React web active cases scaling

rejdi:
#113 implements scaling feature of active cases - all 3 modes.

### DIFF
--- a/web/react-web/src/Content/Graph/Graph.js
+++ b/web/react-web/src/Content/Graph/Graph.js
@@ -1,7 +1,10 @@
 import React, {Component} from 'react';
 import Highcharts from 'highcharts'
 import HighchartsReact from 'highcharts-react-official'
-import {AXES_LINEAR, AXES_LOG, AXES_LOG_LOG} from "../../Commons/sharedObjects";
+import {
+    AXES_LINEAR, AXES_LOG, AXES_LOG_LOG, SCALING_ABSOLUTE, SCALING_PER_CAPITA,
+    SCALING_SAME_PEAK
+} from "../../Commons/sharedObjects";
 
 const colors = ["#7cb5ec", "#434348", "#90ed7d", "#f7a35c", "#8085e9", "#f15c80", "#e4d354", "#2b908f", "#f45b5b", "#91e8e1"];
 
@@ -34,6 +37,9 @@ const predefinedOptions = {
         type: 'linear',
         title: {
             text: 'Active cases'
+        },
+        labels: {
+            format: '{value}'
         }
     },
     legend: {
@@ -96,6 +102,7 @@ class Graph extends Component {
         const {
             axesType,
             isXAxisRelative,
+            dataScaling
         } = options.options;
 
         const finalOptions = {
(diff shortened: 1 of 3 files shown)


web/react-web/src/Content/Graph/Graph.js
```diff
@@ -142,9 +159,32 @@ class Graph extends Component {
...
             }
 
+            let yRatio = 1.0;
+            switch (dataScaling) {
+                case SCALING_SAME_PEAK:
+                    let maxValue = one.max_value;
+                    if (one.type !== 'prediction') {
+                        maxValue = Math.max(...one.values);
+                    }
+                    yRatio = maxPeakValue / maxValue / maxPeakValue * 100.0;
```
lukipuki: Something is off here, the `maxPeakValu` cancels out and `yRatio` is just `maxValue` multiplied by 100?
```suggestion
                    yRatio = maxValue * 100.0;
```

web/react-web/src/Content/Graph/Graph.js
```diff
@@ -142,9 +159,32 @@ class Graph extends Component {
                 name = `Active cases for ${one.long_name}`;
             }
 
+            let yRatio = 1.0;
+            switch (dataScaling) {
+                case SCALING_SAME_PEAK:
+                    let maxValue = one.max_value;
+                    if (one.type !== 'prediction') {
+                        maxValue = Math.max(...one.values);
```
lukipuki, line 157: `one.values` are ative cases or predictions?

rejdi: Both, it just depends on type. For predictions, we have maxValue precalculated. For active cases, we don't. I was already thinking on precalculation of this value for active cases on python side, but it would be another PR with greater set of optimizations.

web/react-web/src/Content/Graph/Graph.js
```diff
@@ -142,9 +149,32 @@ class Graph extends Component {
...
+                case SCALING_ABSOLUTE:
+                default:
+                    finalOptions.yAxis.title.text = 'Active cases';
+                    finalOptions.yAxis.labels.format = '{value}';
+                    break;
+            }
 
             const lineData = one.date_list.map((date, index) => {
-                const y = one.values[index];
+                const y = one.values[index] * yRatio;
```
lukipuki, line 177: Is this the place where the value gets rounded to an integer? For countries with small numbers it would be better not showing rounded numbers but with perhaps one decimal place.

Alternatively bump the scale to cases per 1 million and we might not need any rounding.

web/react-web/src/Content/Graph/Graph.js
```diff
@@ -142,9 +149,32 @@ class Graph extends Component {
...
+                    if (one.type !== 'prediction') {
+                        maxValue = Math.max(...one.values);
+                    }
+                    yRatio = 1.0 / maxValue * 100.0;
+
+                    finalOptions.yAxis.title.text = 'Active cases %';
+                    finalOptions.yAxis.labels.format = '{value}%';
+                    break;
+                case SCALING_PER_CAPITA:
+                    yRatio = 1.0 / (one.population / 100000.0);
```
lukipuki: From a usability perspective, 100000 is perhaps not the best. People know the population in millions, so it might be best to show cases per million (as wroldometers.info is doing https://www.worldometers.info/coronavirus/#countries). That way you can easily multiply it back in your head if needed.

Also this has the nice sideffect of 10-times larger numbers, so we get more precision. This also relates to my other comment regarding precision.

web/react-web/src/Content/Graph/Graph.js
rejdi: Rounding is actualy done when rendering tooltip. I'll change it to 1 000 000.

web/react-web/src/Content/Graph/Graph.js
rejdi: Done.

lukipuki (approved):

PR #127, opened 2020-05-16
Add Google Analytics support

lukipuki:
Right now only the /daily endpoint works for a mysterious reason.

Perhaps I'm missing something obvious, but I don't see how `/daily` is different from others.


lukipuki:
Fixed the problem with `/single` and `/daily`, it's a known Dash issue: https://github.com/plotly/dash/issues/214

python/covid_web/covid_web/index.html
```diff
@@ -5,6 +5,17 @@
...
+    <!-- Global site tag (gtag.js) - Google Analytics -->
+    <script async src="https://www.googletagmanager.com/gtag/js?id=UA-676869-6"></script>
+    <script>
+      window.dataLayer = window.dataLayer || [];
+      function gtag(){dataLayer.push(arguments);}
+      gtag('js', new Date());
+
+      gtag('config', 'UA-676869-6');
+    </script>
+
```
rejdi, line 18: Could you add this to react-web as well?

python/covid_web/covid_web/server.py
```diff
@@ -32,6 +34,24 @@ def run_server(data_dir: Path, prediction_dir: Path) -> None:
...
+        "tid": GA_TRACKING_ID,  # Tracking ID / Property ID.
+        "cid": 47,
+        "t": "pageview",
+        "dp": f"/covid19/{path}",
+        "dh": "graphs.lukipuki.sk",
+        "ua": request.headers.get("User-Agent"),
+    }
+
+    response = requests.post("https://www.google-analytics.com/collect", data=data)
+    response.raise_for_status()
```
rejdi, line 52: Do we really need to add also a python code for tracking? Isn't included JS in HTML enough?

lukipuki: If I were to choose, I'd choose the Python part and drop JS from `index.html`. The `index.html` script is only for the root page, which no one visits. I put it there mostly to test the functionality, as the Python version wasn't working.

python/covid_web/covid_web/index.html
rejdi: Added.

rejdi (approved):

PR #128, opened 2020-05-16
Bigger refactor to better clarify responsibility.

rejdi:
Greatly improved performance for many selected countries.

Sorry guys for this bigger one, but it was necessary and I couldn't break it down into smaller pieces.


web/react-web/src/Content/Graph/Graph.js
```diff
@@ -69,6 +69,16 @@ function xaxis_text(x, isXAxisRelative) {
     }
 }
 
+function generateNumber(value) {
```
lukipuki: Is this a random number generator? Or a hash calculation?

It seems it's the latter, so please call the function `calculateHash` or similar.

web/react-web/src/Commons/functions.js
```diff
@@ -1 +1 @@
-export const isSetsEqual = (a, b) => a.size === b.size && [...a].every(value => b.has(value));
\ No newline at end of file
+export const isSetsEqual = (a, b) => a.size === b.size && [...a].every(value => b.has(value));
```
lukipuki:
```suggestion
export const areSetsEqual = (a, b) => a.size === b.size && [...a].every(value => b.has(value));
```

lukipuki (approved):
Can't review too much of the JS but if it works, ship it!

web/react-web/src/Content/Graph/Graph.js
rejdi: Applied.

web/react-web/src/Commons/functions.js
rejdi: Applied.

PR #133, opened 2020-05-18
New web slider usability fixes.

rejdi:
Add rc-slider workaround for keyboard navigation bug.
Use own slider tooltip, so it's always shown correctly.
Autoselect latest prediction when nothing is selected.

### DIFF
--- a/web/react-web/public/index.html
+++ b/web/react-web/public/index.html
@@ -8,7 +8,7 @@
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
     <meta
       name="description"
-      content="Web site created using create-react-app"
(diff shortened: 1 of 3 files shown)


lukipuki:
Slider can be navigated using keyboard? What should I press?

rejdi:
For now, only Home and End works. Slider itself handle arrow keys and PgUp, PgDown, but it's buggy and may lead to invalid states (press once or twice left arrow key and you'll lose prediction, because invalid value is send from slider component). I'm willing to fix&PR into original component or fork it, but for now it's better to not have this feature.

EDIT: slider must have focus. That gives me an idea to handle arrow keys everywhere on page and move with slider&predictions accordingly.

lukipuki (approved):

PR #135, opened 2020-05-31
Added link to new web.

rejdi:


python/covid_web/covid_web/index.html
```diff
@@ -21,6 +21,9 @@
     <h1>COVID-19 predictions of Katarína Boďová and Richard Kollár</h1>
     Dashboards with predictions:
     <ul>
+      <li>
+        <a href="/covid19/predictions/overview/">predictions overview</a>
```
lukipuki: I think this is not descriptive enough. And the other 3 links are not clear either. I propose something like this, but feel free to expand it:

- overview of all countries in a single graph

- single country predictions with a slider

- single country predictions

- overview of all countries on one long page

rejdi: My proposal:
- predictions overview with country selection and timeline
- single country predictions with timeline
- single country with last 5 predictions in one graph
- latest predictions for all countries

lukipuki: Using "timeline" here is incorrect, see e.g. https://www.google.com/search?tbm=isch&source=hp&q=timeline&oq=timeline&sclient=img I'd keep "slider" there.

"latest predictions" is also incorrect, since we allow going back a few weeks, which isn't really "latest". Similarly "last predictions" is not correct. 

- predictions overview with country selection and a slider
- single country predictions with slider
- single country with multiple predictions in one graph
- overview of all countries on one long page

rejdi: Applied.

lukipuki (approved):

PR #136, opened 2020-08-06
Add data until June 30

lukipuki:
Predictions are calculated for 10 weeks, so from April 22 to June 30.

There's something wrong with Jordan, so I dropped it.


miskosz (commented):
LGTM ;)

PR #137, opened 2020-11-09
Allow specifying start and end date for data

lukipuki:


python/setup_environment.sh
```diff
@@ -2,7 +2,7 @@
 set -euo pipefail
 
 rm -rf .venv
-python3.7 -m venv .venv
+python3.7 -m venv --system-site-packages .venv
```
miskosz, line 5: Why was this needed?

miskosz (commented):
LGTM :)

python/setup_environment.sh
lukipuki: This is more for my convenience, I don't want to fill my disk with unnecessary copies of packages :slightly_smiling_face:  Maybe this should have been hidden by an option of the script. I can add a flag `--use-system-packages` or similar if you want.

